PR #78, opened 2014-12-28
Improvements for nimble build.

minciue:
- `nimble build` now compiles in debug mode while `nimble install` compiles in release mode.
- add a `builddir` option to .nimble files to specify where built binaries/libs should go.


gradha:
Why is there a need for having ``nimbe build`` produce a debug binary? I would actually prefer if nimble had a switch to always install a debug copy side by side with the release binary. Where ``binaryName.exe`` is installed, generate a ``binaryNameD.exe``. Then if something fails for a user and they get a sigsev, they can run the same command with a ``D`` appended and produce a useful bug report.

dom96:
Sorry for the delay.

PR #82, opened 2014-12-31
Add command to list installed packages.

hiteshjasani:
A package manager should have a way to list installed packages, so I added one.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -4,6 +4,8 @@
 import httpclient, parseopt, os, strutils, osproc, pegs, tables, parseutils,
        strtabs, json, algorithm, sets
 
+from terminal import nil
+
 from sequtils import toSeq
 
 import nimblepkg/packageinfo, nimblepkg/version, nimblepkg/tools,
@@ -22,11 +24,11 @@ type
 
   TActionType = enum
     ActionNil, ActionUpdate, ActionInit, ActionInstall, ActionSearch,
-    ActionList, ActionBuild, ActionPath, ActionUninstall
+    ActionList, ActionBuild, ActionPath, ActionUninstall, ActionListInstalled
 
   TAction = object
     case typ: TActionType
-    of ActionNil, ActionList, ActionBuild: nil
+    of ActionNil, ActionList, ActionBuild, ActionListInstalled: nil
     of ActionUpdate:
       optionalURL: string # Overrides default package list.
     of ActionInstall, ActionPath, ActionUninstall:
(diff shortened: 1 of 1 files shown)


src/nimble.nim
```diff
@@ -663,6 +668,29 @@ proc list(options: TOptions) =
       echoPackageVersions(pkg)
     echo(" ")
 
+proc c_printf(frmt: cstring) {.importc: "printf", header: "<stdio.h>",
```
dom96: Use Nim's string formatting features please.

src/nimble.nim
```diff
@@ -663,6 +668,29 @@ proc list(options: TOptions) =
...
+  varargs.}
+
+proc listInstalled(options: TOptions) =
+  let pkgsDir = options.getNimbleDir() / "pkgs"
+  if not existsDir(pkgsDir):
+    raise newException(ENimble, "Nimble dir not installed.")
```
dom96: Change that to: ``raise newException(ENimble, "Could not find Nimble pkgs dir: " & pkgsDir)``

src/nimble.nim
```diff
@@ -55,6 +57,7 @@ Commands:
   search       [--ver] pkg/tag    Searches for a specified package. Search is
                                   performed by tag and by name.
   list         [--ver]            Lists all packages.
+  installed                       Lists all installed packages and versions.
```
dom96: I would prefer a flag for the ``list`` subcommand. ``--installed`` perhaps.

src/nimble.nim
```diff
@@ -663,6 +668,29 @@ proc list(options: TOptions) =
...
 
+proc c_printf(frmt: cstring) {.importc: "printf", header: "<stdio.h>",
+  varargs.}
+
+proc listInstalled(options: TOptions) =
+  let pkgsDir = options.getNimbleDir() / "pkgs"
+  if not existsDir(pkgsDir):
+    raise newException(ENimble, "Nimble dir not installed.")
+  var h = initTable[string, seq[string]]()
+  for kind, path in walkDir(pkgsDir):
```
dom96: You should use ``packageinfo.getInstalledPkgs`` here.

hiteshjasani:
A couple of thoughts:
* From what I can tell, Nim's string formatting features are nice but do not match up with the features and terseness of `printf` quite yet.  I have some ideas about how to improve the string formatting, but that would need to happen outside of nimble.  The end result is that I've removed the pretty printing of package names and versions.
* I moved the option to under the `list` command using either `-i` or `--installed`.  I wish the option parser in nimble were more advanced to avoid using up the `-i` option for only the list command.  But it's not worth reimplementing the option parser just to add this simple command.  On the flip side, being able to list installed packages should be a very short command, which `nimble list -i` accomplishes so I'm happy with the result.

dom96:
Brilliant thank you!

Once something else starts using ``-i`` we can improve the situation.

PR #111, opened 2015-03-30
Fix #110 : use '^' rather than '-' for list negative indexes

### LINKED ISSUE #110: Installer: use '^' instead of '-'; negative indexing is obsolete
OS: Debian Jessie RC2: 3.16.0-4-amd64
~/github/nimble$ nim c -r src/nimble install
...
lib/pure/osproc.nim(779, 6) Warning: raiseOSError is deprecated [Deprecated]
lib/pure/osproc.nim(701, 9) Hint: 'osproc.startProcessAuxSpawn(data: TStartProcessData)' is declared but not used [XDeclaredButNotUsed]
Hint: pegs [Processing]
Hint: unicode [Processing]
Hint: json [Processing]
Hint: lexbase [Processing]
Hint: algorithm [Processing]
Hint: sets [Processing]
lib/pure/collections/sets.nim(801, 5) Hint: 'sets.testModule()' is declared but not used [XDeclaredButNotUsed]
Hint: sequtils [Processing]
Hint: packageinfo [Processing]
Hint: parsecfg [Processing]
Hint: version [Processing]
**/home/jdunlap/github/nimble/src/nimblepkg/version.nim(142, 21) Error: use '^' instead of '-'; negative indexing is obsolete**

JCavallo:
Just following the compiler's hint.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -171,7 +171,7 @@ proc parseCmdLine(): Options =
           # Parse pkg@verRange
           if '@' in key:
             let i = find(key, '@')
-            let pkgTup = (key[0 .. i-1], key[i+1 .. -1].parseVersionRange())
+            let pkgTup = (key[0 .. i-1], key[i+1 .. ^1].parseVersionRange())
             result.action.packages.add(pkgTup)
           else:
             result.action.packages.add((key, VersionRange(kind: verAny)))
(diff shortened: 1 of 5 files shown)


dom96:
Please use ``str.len-1`` instead of ``^1``

saml:
looks like these are the same:

```nim
arr[0 .. arr.len() - N]
arr[0 .. ^N]
```

dom96:
They are but the latter won't work with an older compiler.

JCavallo:
Done.

Tests do not run on my machine, but that may be because of some missing configuration.

saml:
:+1:

Issue #115, opened 2015-04-21
nimble update failures


bougyman:
On void linux, Nim Compiler Version 0.10.2 (2015-02-18) [Linux: amd64], fresh install of the master branch of nimble, running `nimble update` returns

```
bougyman@xena ~/g/nimble % nimble update
Downloading package list from https://github.com/nim-lang/packages/raw/master/packages.json
Error: unhandled exception: 404 Not Found [HttpRequestError]
bougyman@xena ~/g/nimble % nimble update https://github.com/nim-lang/packages/raw/master/packages.json
Downloading package list from https://github.com/nim-lang/packages/raw/master/packages.json
Error: unhandled exception: 404 Not Found [HttpRequestError]
bougyman@xena ~/g/nimble % nimble update https://raw.githubusercontent.com/nim-lang/packages/master/packages.json
Downloading package list from https://raw.githubusercontent.com/nim-lang/packages/master/packages.json
Error: unhandled exception: 404 Not Found [HttpRequestError]
bougyman@xena ~/g/nimble % wget https://raw.githubusercontent.com/nim-lang/packages/master/packages.json
--2015-04-21 12:59:12--  https://raw.githubusercontent.com/nim-lang/packages/master/packages.json
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 199.27.78.133
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|199.27.78.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 42603 (42K) [text/plain]
Saving to: ‘packages.json’

packages.json                                               100%[============================================================================================================================================>]  41.60K  --.-KB/s   in 0.04s  

2015-04-21 12:59:12 (1.07 MB/s) - ‘packages.json’ saved [42603/42603]
```

Any ideas?

bougyman:
This may have to do with https, I downloaded it with wget (above) and moved it to my own web server in http and I do not have the issue:

```
% nimble update http://dev.bougyman.com/packages.json
Downloading package list from http://dev.bougyman.com/packages.json
Done.
```

Any chance this is not really a 404 but an ssl error underneath?

boydgreenfield:
I'm having/reproducing the same issue. Looks likely SSL related to me as well.

Here's the bad request that `get` in `downloadFile` is getting:

```
[...]

PR #117, opened 2015-05-01
Fix default packages URL

grncdr:
It seems as though github has moved raw file hosting to a different domain


grncdr:
Hm, actually, this doesn't appear to be the issue, even after this change I am getting 404's when running `nimble update`

dom96:
This perhaps? #115

grncdr:
Yeah I'm pretty sure that's the issue I'm running into. Being a new nim user, I'm not really able to debug it further.

PR #118, opened 2015-05-02
Fix #114: Overrides the nimbleDir option with value from command line

### LINKED ISSUE #114: Implement --dir option to override where nimbleDir
For reasoning check discussion on IRC at about 6pm: http://irclogs.nim-lang.org/11-04-2015.html

ClementJnc:


dom96:
Please call the option ``--nimbleDir``.

PR #147, opened 2015-07-08
Fix #96 Ask for a value for every Package fields during 'nimble init'

### LINKED ISSUE #96: More options for nimble init
Right now `nimble init` only sets the name, nothing else. Should have more options for everything else.

dgellow:
The result is like this:


dom96:
Sorry, ended up implementing this a bit differently.

dgellow:
@dom96 :ok_hand:

PR #149, opened 2015-07-21
Fixed chcp call bug on windows xp

vegansk:
On Windows XP wrapper script fails on chcp call. See details and workaround http://stackoverflow.com/questions/2182568/batch-script-is-not-executed-if-chcp-was-called


dom96:
Hopefully this will work in newer Windows versions too.

vegansk:
@dom96, unfortunatly, it breaks down things on Vista and newer. I made another PR: https://github.com/nim-lang/nimble/pull/150. Or do I need to fix this?

PR #152, opened 2015-07-26
Added local path support with "local" arg (i.e. nimble install local …

onionhammer:
…<pkgname>)


dom96:
Apologies for neglecting this PR for so long. By now I think that the ``--nimbleDir`` flag implements this functionality in a more flexible way, as such I will be closing this.

PR #155, opened 2015-08-18
implemented 'nimble publish'

Araq:


src/nimblepkg/version.nim
```diff
@@ -29,6 +29,9 @@ type
     of verAny:
       nil
 
+  ## Tuple containing package name and version range.
+  PkgTuple* = tuple[name: string, ver: VersionRange]
```
dom96: Why did you move this here?

src/nimblepkg/publish.nim
```diff
@@ -0,0 +1,191 @@
...
+
+proc createFork(a: Auth) =
+  discard postContent("https://api.github.com/repos/nim-lang/packages/forks",
+      extraHeaders=("Authorization: Basic $1\c\L" % a.token) &
+        "Content-Type: application/x-www-form-urlencoded\c\L" &
+        "Accept: */*\c\L")
+
+proc createPullRequest(a: Auth; packageName: string) =
+  echo "creating PR"
+  discard postContent("https://api.github.com/repos/nim-lang/packages/pulls",
```
dom96: Check what the result is here.

src/nimblepkg/publish.nim
```diff
@@ -0,0 +1,191 @@
...
+      extraHeaders=("Authorization: Basic $1\c\L" % a.token) &
```
dom96: I doubt this method of auth will work if I have got two-factor authentication enabled. Could we not do oauth?

src/nimblepkg/publish.nim
dom96: Also, we're back to depending on OpenSSL :(

src/nimblepkg/publish.nim
```diff
@@ -0,0 +1,191 @@
...
+proc getPackageOriginUrl(a: Auth): string =
+  ## Adds 'user:pw' to the URL so that the user is not asked *again* for it.
+  ## We need this for 'git push'.
+  let (output, exitCode) = doCmdEx("git config --get remote.origin.url")
+  result = "origin"
+  if exitCode == 0:
+    result = output.string.strip
+    if result.endsWith(".git"): result.setLen(result.len - 4)
+    if result.startsWith("https://"):
+      result = "https://" & a.user & ':' & a.pw & '@' &
```
dom96: Use the URI module here.

src/nimblepkg/publish.nim
```diff
@@ -0,0 +1,191 @@
...
+          result["https://".len .. ^1]
+
+proc publish*(p: PackageInfo) =
+  ## Publishes the package p.
+  let auth = getGithubAuth()
+  let parent = os.getCurrentDir().parentDir()
```
dom96: This should probably use getTempDir.

src/nimblepkg/publish.nim
```diff
@@ -0,0 +1,191 @@
...
+  var pkgsDir = parent / "nimble-packages-fork"
+  if not forkExists(auth):
+    createFork(auth)
+    echo "waiting 10s to let Github create a fork ..."
```
dom96: This sounds insanely unreliable.

dom96:
In general, very good and very cool, but would be nice to get oauth support. Would also be safer.

src/nimblepkg/version.nim
Araq: Perhaps not necessary now, but necessary for the upcoming PRs. Also I think it does make sense here too.

src/nimblepkg/publish.nim
Araq: True but works in practice and github recommends 5 seconds. But on a second though, we can busy wait for 1 minute, checking from time to time 'forkExists'. But I think it's good enough for now. Let people report problems.

src/nimblepkg/publish.nim
Araq: Nope, we don't want the fork to be in a temporary directory.

src/nimblepkg/publish.nim
Araq: I don't like the dependency and it's been tested with this code. Testing this cannot be done effectively with unit tests and took as much time as writing it.

src/nimblepkg/publish.nim
Araq: Yes, but Nimble already requires '-d:ssl'.

Araq: Also, I don't have to check. It raises an exception if it fails and the output is a huge JSON full of info I don't care about.

src/nimblepkg/publish.nim
Araq: Again, tested it this way and it is MUCH better than nothing. Let others complain and improve it.

src/nimblepkg/publish.nim
dom96: Why?

Araq: Because we don't want to 'git clone' all the time.

PR #156, opened 2015-09-03
Add NimScript support to Nimble

Araq:


src/nimble.nim
```diff
@@ -112,21 +117,21 @@ proc writeVersion() =
       [nimbleVersion, CompileDate, CompileTime])
   quit(QuitSuccess)
 
-proc getNimbleDir(options: Options): string =
-  options.config.nimbleDir
+proc getNimbleDir(opt: Options): string =
```
dom96: why?

Araq: Because I found it slightly confusing with the compiler/options import.

andreaferretti:
Is this being discussed somewhere? It would be great if this was merged, so that one could use a single `config.nims` to write tasks and dependencies!

andreaferretti:
Are there any news about this PR?

dom96:
Yeah. I'm working on it. Needs some adjustments.

dom96:
Note to self: ``WARNING: No .nimble file found for /Users/dom/.nimble/pkgs/jester-1.0.0`` need to fix this if installed with a nimscript.

dom96:
Current blockers:
  * The bug above.
  * ``setCommand "foo"``
  * The ``.nimble`` extension should be reused for nimscript files, instead of using ``.nims``.

andreaferretti:
Thaank you, nice Christmas present! ;-)

dom96:
No problem. Please test it if you can :)

There are still some issues to work out of course. Some which I am aware of, and probably many others which I am not.

PR #170, opened 2015-12-23
Add Bash completion

FedericoCeratto:


dom96:
Nice!

nimble.bash-completion
```diff
@@ -0,0 +1,40 @@
+# vim: filetype=sh
+
+_nimble()
+{
+    local cur=${COMP_WORDS[COMP_CWORD]}
+    local prev=${COMP_WORDS[COMP_CWORD-1]}
+    COMPREPLY=()
+    _init_completion || return
```
oderwat: This line gives an error on my system (OSX). I simply removed it.. Is this working on Linux?

dom96: @FedericoCeratto ^

FedericoCeratto:
@oderwat it does.

oderwat:
Well Macs bash completion doesn't have this for the installed version of the completion. One could add a minimal version which looks like this:

```
__my_init_completion()
{
    COMPREPLY=()
    _get_comp_words_by_ref cur prev words cword
}
```

But afais it is not at all needed for your completion code and works without it. Why not remove it or just call it if it is available?

PR #174, opened 2015-12-24
Fixed compilation

yglukhov:
This fixed nimble compilation which otherwise fails with
```
nimblepkg/nimscriptsupport.nim(15, 43) Error: undeclared identifier: '$'
```


dom96:
Forgot to push my `$` to Nim's repo. But better to get rid of this anyway :)

Issue #181, opened 2015-12-30
Ability to import NimScript modules in tasks


dom96:
This may need to be done at runtime, because placing an import statement at the top of a Nimble file will cause the module to be imported each time the Nimble file is evaluated.

genotrance:
So with #635, nimble files are just executed with `nim e`. Assuming that PR goes through, is there something extra needed within nimble? 

Nim only allows imports at the top level so seems like this should be tracked there.

dom96:
I think I had a fancy mechanism in mind that would allow importing modules that are perhaps offered by other packages, this would mean that the imports would need to be in the tasks and that the tasks would only be executable when those packages are installed. It's possible this isn't really all that useful.

genotrance:
I propose closing this.

dom96:
I'll demote this, but I'd like to keep it as a potential feature.

PR #185, opened 2016-01-01
[RFC] Added nimbleDependenciesInstalled

yglukhov:
This is a proposition to define `nimbleDependenciesInstalled` on the task phases. This solves a problem of importing modules which are dependencies to nimble file. The solution is not super convenient, but it's better than nothing, IMHO.

Related to #181.


yglukhov:
I guess i'll close it for now, because I find nake much more convernient anyway.

dom96:
That's a pity, please create an issue with whatever needs to be added to Nimble to make it more convenient :)

PR #188, opened 2016-01-05
Updated README syntax

aidansteele:
It appears the `bin` directive now expects a seq rather than a string. Otherwise I get this error:

> Error: type mismatch: got (string) but expected 'seq[string]'. [NimbleError]

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -489,7 +489,7 @@ A package is automatically a binary package as soon as it sets at least one
 ``bin`` value, like so:
 
 ```ini
-bin = "main"
+bin = @["main"]
 ```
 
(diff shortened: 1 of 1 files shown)


dom96:
Thank you! But could you instead add a note below that example? Some people may still be using/expecting the old ini-format and get confused. Something like ``**Note**: For NimScript .nimble files make sure to use bin = @[...] `` or maybe a comment inside the example.

dom96:
Ping. It's cool if you don't to do that, just close if so.

aidansteele:
My sincere apologies for not doing this sooner. I've now made the recommended change. Please let me know if there is anything else I can do to help.

dom96:
Thank you :)

PR #195, opened 2016-01-11
Fix for the Bash completion on OSX.

oderwat:
On OSX there is no `_init_completions` even if `brew install bash-completion` was installed. The fix is to not call it if it is not defined. IMHO it is never needed, but I may be wrong here.


dom96:
@FedericoCeratto think this is okay to merge?

FedericoCeratto:
Yup!

PR #196, opened 2016-01-11
Smarter uninstall Bash completion.

oderwat:
Instead of offering just the base names of the packages the uninstall parameter completion now lists all installed versions. This makes it easy to uninstall whats not needed anymore and helps remembering the
syntax for this task.

Example:

```
$ nimble uninstall nake[TAB]@1.[TAB]
nake@1.4  nake@1.6  nake@1.8
$ nimble uninstall nake@1.4
```


dom96:
Awesome!

PR #197, opened 2016-01-11
This change sorts the list of the installed modules by name.

oderwat:
I thought that it is easier to skim the installed modules if the list is sortet by package name.

P.S.: I also added some files to `.gitignore` not sure if that needs to be an extra PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,8 @@ nimcache/
 /src/babel
 /src/nimble
 /tests/tester
+
+# executables from test and build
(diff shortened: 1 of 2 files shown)


dom96:
Thank you! Just so you know, I made a small tweak to this PR (just me nitpicking).

oderwat:
I was wondering if it is better to not use auto there but thought the compiler will just figure it out perfectly and not generate different code for that.

Issue #198, opened 2016-01-12
undeclared identifier 'map' when compiling


JorySchossau:
I get this error compiling from the latest master nim, and devel nim. Not something I've observed before updating nim and nimble repos.

dom96:
I don't think this is a bug.

From readme:

>The latest version of Nimble (0.7.0) requires a version of Nim greater than 0.12.0.

JorySchossau:
Thanks; I must've missed something trying with devel last time.

dom96:
Please reopen if you can't get it to compile with devel.

PR #199, opened 2016-01-12
Update packageinfo.nim

JorySchossau:
#198
This fixes a problem I get compiling nimble "undeclared identifier 'map'".
Fixes bug(?) #198


dom96:
Closing. For reason see #198.

PR #208, opened 2016-02-17
Fixed permission issue when installing package

SSPkrolik:
If there are executables in Git repo of the project with set FS permission bits, files must be copied via copyFileWithPermissions()


dom96:
Thank you! :)

PR #214, opened 2016-02-25
Fix for Issue 204

Jeff-Ciesielski:
Github + https + trailing slash in URL causes failed checkouts / remote
tag fetches.  The issue appears to be indemic to github.com, not git itself

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -492,7 +492,7 @@ proc downloadPkg(url: string, verRange: VersionRange,
   # Fixes issue #204
   # github + https + trailing url slash causes a
   # checkout/ls-remote to fail with Repository not found
-  if modUrl.contains("https://github.com") and modUrl.endswith("/"):
+  if modUrl.contains("github.com") and modUrl.endswith("/"):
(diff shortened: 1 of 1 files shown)


src/nimble.nim
```diff
@@ -484,10 +484,17 @@ proc downloadPkg(url: string, verRange: VersionRange,
...
-  let modUrl =
+  var modUrl =
     if url.startsWith("git://") and options.config.cloneUsingHttps:
       "https://" & url[6 .. ^1]
     else: url
+
+  # Fixes issue #204
+  # github + https + trailing url slash causes a
+  # checkout/ls-remote to fail with Repository not found
+  if modUrl.contains("https://github.com") and modUrl.endswith("/"):
```
dom96: what about ``git://github.com``? Might be better to just check for ``github.com``.

Jeff-Ciesielski: I thought about that, but my concern was a package that (for some odd reason) included github.com in it's url (which I suppose might overreaching a little bit).

Perhaps `://github.com'` is a happy medium?

dom96:
Thanks :)

PR #218, opened 2016-03-18
Fixed cygwin stub

yglukhov:
Fixed broken cygwin stub.
1. Use cygpath to convert windows path to unix path.
2. Use linefeed at the end. Cygwin doesnt like windows line endings.
before:
```
"C:\Users\boost\.nimble\pkgs\nimble-0.7.0\nimble.exe" "$@"
```
after:
```
"$(cygpath --unix 'C:\Users\boost\.nimble\pkgs\nimble-0.7.0\nimble.exe')" "$@"
```


dom96:
Love it, thanks!

PR #223, opened 2016-04-06
Fixed FD leak.

yglukhov:
A nasty one. Took half a day to track down =)

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -232,7 +232,7 @@ proc removeRevDep(options: Options, pkg: PackageInfo) =
       for key, val in options.nimbleData["reverseDeps"]:
         options.remove(pkg, depTup, val)
     else:
-      let thisDep = options.nimbleData["reverseDeps"][depTup.name]
+      let thisDep = options.nimbleData{"reverseDeps", depTup.name}
       if thisDep.isNil: continue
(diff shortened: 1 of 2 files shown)


dom96:
Interesting. How did this manifest itself? Surely the FDs are freed when the Nimble process quits anyway?

dom96:
Also, tests failed, could be because of a recent Nim change though.

yglukhov:
It could be reproduced on 1 mac of 3. Doing nimble install of a package with bin, nim suddenly failed (cannot open system/inclrtl.nim). The bug seemed very obscure to say the least. I've logged errno in open function, and it said 24 (out of file descriptors). So we've searched for the bug in nim at first, then nim lib, and eventually found that fd number is continuosly increasing when running nim through nimble.

dom96:
I see. Sorry about that :)

Nim really needs a ``withFile`` template.

yglukhov:
No worries, i'm really happy i found the problem =)
`withFile` wouldn't really help in this case because CfgParser takes the ownership of the file.

yglukhov:
But proper RAII would probably help =)

yglukhov:
Fixed the tests

dom96:
hah, glad my test suite catched that :D

Issue #190, opened 2016-01-07
Package names (still?) not style insensitive


dom96:
`nimble install ringbuffer` fails, but `nimble install RingBuffer` works.

dom96:
This issue has been fixed.

PR #225, opened 2016-04-10
Package names are case insensitive (#190)

aidansteele:
Issue #190 says **style** insensitive, but the example only highlights case sensitivity. This is a fix for case sensitivity. Are package names meant to be style insensitive in the same way that identifiers are in the Nim language? If so, I can update the PR.


dom96:
Yes, they are. Please use ``strutils.normalise``.

aidansteele:
Change made :)

dom96:
Thank you :)

PR #228, opened 2016-05-04
add critical install step to README

alphashuro:
Added critical unix/mac installation step to the readme github install instructions. Without it, the install fails on mac and unix.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -48,6 +48,7 @@ the following commands to clone nimble, compile it and then install it.
     git clone https://github.com/nim-lang/nimble.git
     cd nimble
     git clone -b v0.13.0 --depth 1 https://github.com/nim-lang/nim vendor/nim
+    ln -s vendor/nim/lib .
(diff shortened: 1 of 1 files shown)


dom96:
This shouldn't be necessary, because of this: https://github.com/nim-lang/nimble/blob/master/src/nimble.nim.cfg#L4

dom96:
Any chance you could investigate why that doesn't work?

alphashuro:
Okay I will take a look at it

dom96:
Thanks :)

dom96:
Any luck?

alphashuro:
not yet, just started at a new job so time has been a bit scarce, but I had some leads. I will take another look this weekend?

dom96:
Awesome. No worries if you don't get a chance! :)

dom96:
Hopefully won't be a problem anymore as Nimble is bundled with Nim now.

PR #229, opened 2016-05-19
Add a Troubleshooting section

k0pernicus:
For the moment, it contains only the problem for the SSL support. But it can be great to list recurrent problems/questions associated with ```nimble```.


dom96:
👍

PR #232, opened 2016-05-23
Dependencies only installation

Tormund:
Add dependencies only installation by flag -d


dom96:
Beautiful PR. Thank you!

PR #242, opened 2016-07-06
Error: cannot open 'compiler/ast'

juanfra684:
In the commit e95df37ebedaf7b530df8c8153762f9f8e8b1992, the command `git clone -b v0.13.0 --depth 1 https://github.com/nim-lang/nim vendor/nim` was deleted from the readme. I can't install nimble on OpenBSD or Arch Linux without that command. The build process shows this error:
```
[...]
Hint: config [Processing]                                                                        
Hint: download [Processing]                                                                      
Hint: packageparser [Processing]                                                                 
Hint: nimscriptsupport [Processing]                                                              
/tmp/nimble/src/nimblepkg/nimscriptsupport.nim(8, 11) Error: cannot open 'compiler/ast'
```
This PR adds back the missing command to the readme with the version updated.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -83,6 +83,7 @@ the following commands to clone nimble, compile it and then install it.
 
     git clone https://github.com/nim-lang/nimble.git
     cd nimble
+    git clone -b v0.14.2 --depth 1 https://github.com/nim-lang/nim vendor/nim
     nim -d:release c -r src/nimble -y install
 
 After these steps are completed successfully, nimble will be installed


dom96:
We should instead work out why you cannot install Nimble without that command. Where does the Nim installation reside on your system?

juanfra684:
OpenBSD: `/usr/local/lib/nim` (https://gitlab.com/juanfra684/openbsd-ports/blob/master/lang/nim/pkg/PLIST). Arch Linux: `/usr/lib/nim` (https://www.archlinux.org/packages/community/x86_64/nim/files/).

dom96:
Okay. So I think I know why, the Arch and OpenBSD package doesn't contain the compiler sources. It should contain them. But I don't think this is the packager's fault. I created the relevant Nim issue for this: https://github.com/nim-lang/Nim/issues/4866

dom96:
Thank you for taking the time to create this btw!

Araq:
You build Nimble with Nim now or rely on your package manager to provide both for you. There is no bug here.

PR #249, opened 2016-08-31
Add a postinstall hook

stisa:
Adds the possibility to run a hook when nimble installs a package.
The hook is set by adding a block in the `.nimble` file:
```nim
postinstall:
  echo "Post install hook"
```
This is useful for example when installing packages that use [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) or when a binary needs to register with application, run tests on install, run custom scripts etc.

I used execTask as a base and modified it to run the hook, so this may not be the best/cleanest way of implementing. Any feedback is appreciated.


stisa:
Closing as I made a mess, I'll try to find a nicer way of doing this.

PR #252, opened 2016-09-19
Add binDir checking/auto creation.  Fixes #215

### LINKED ISSUE #215: nimble install fails on a fresh clone of a repo
When using the `srcDir` and `binDir` options in a `.nimble` file, if you run `nimble install` or `nimble build` in the directory and the `binDir` has not yet been created, it fails because it does not check whether the `binDir` exists or not before it begins the compilation.

Here is my `.nimble` for reference:
```
[Package]
name          = "playfair"
version       = "0.1.0"
author        = "Benjamin Clos"
description   = "Playfair encryption implementation"
license       = "MIT"
srcDir        = "src"
binDir        = "lib"
bin           = "playfair"

[Deps]
Requires: "nim >= 0.10.0, docopt"
```

This seems counter-intuitive to me, as generally you don't check your compiled code into git, so it's unlikely that the `binDir` will be present for CI builds or other automated deployments.

for now I am just going to add a `mkdir ./lib` so it doesn't error out, but we should probably check to see if the `binDir` exists and if it doesn't create it in the build process.

Here is an [example build](https://travis-ci.org/ondreian/playfair/builds/114002652#L676) that fails in this manner.

Jeff-Ciesielski:
FWIW: I think perhaps pkgInfo.getOutputDir could use a renaming as it seems to return the output path (including the binary) rather than the output directory.  Could also be that I misunderstand the use of that function though.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -313,6 +313,11 @@ proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], forRelease: bool) =
     let outputOpt = "-o:\"" & pkgInfo.getOutputDir(bin) & "\""
     echo("Building ", pkginfo.name, "/", bin, " using ", pkgInfo.backend,
          " backend...")
+
+    let outputDir = pkgInfo.getOutputDir("")
+    if not existsDir(outputDir):
+      createDir(outputDir)
+
     try:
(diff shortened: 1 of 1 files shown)


dom96:
Yeah, I think you're right. We should rename ``getOutputDir`` but maybe implement a proper ``getOutputDir`` as well.

PR #253, opened 2016-09-21
Store vcs revision in metadata

yglukhov:
Basically subj. The reason behind this feature request is that we need some way to track our package revisions, that change in git many times a day, and manually updating version numbers in nimble files is cumbersome and error prone. For now I just want to test this feature with no public api exposed to see how it goes.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -326,11 +326,14 @@ proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], forRelease: bool) =
       raise newException(BuildFailed, "Build failed for package: " &
                          pkgInfo.name)
 
-proc saveNimbleMeta(pkgDestDir, url: string, filesInstalled: HashSet[string]) =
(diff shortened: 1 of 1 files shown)


src/nimble.nim
```diff
@@ -326,11 +326,14 @@ proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], forRelease: bool) =
...
+proc saveNimbleMeta(pkgDestDir, url, vcsRevision: string, filesInstalled: HashSet[string]) =
   var nimblemeta = %{"url": %url}
-  nimblemeta["files"] = newJArray()
+  if not vcsRevision.isNil:
+    nimblemeta["vcsRevision"] = %vcsRevision
+  let files = newJArray()
+  nimblemeta["files"] = files
   for file in filesInstalled:
-    nimblemeta["files"].add(%changeRoot(pkgDestDir, "", file))
+    files.add(%changeRoot(pkgDestDir, "", file))
```
dom96: Why this change? Just didn't want to write out ``nimblemeta["files"]``?

yglukhov: Yep, also a performance improvement! =)

PR #268, opened 2016-10-31
make Nimble compile with the changed compiler API

Araq:


src/nimblepkg/nimscriptsupport.nim
```diff
@@ -11,7 +11,7 @@ import
   compiler/msgs, compiler/magicsys, compiler/lists
 
 from compiler/scriptconfig import setupVM
-from compiler/idents import getIdent
+from compiler/idents import getIdent, newIdentCache
```
dom96: Won't this fail on older versions?

Araq: Nope. :-) Nim is awesome.

dom96: Hrm. Cool, but that should give a warning at least ;)

Issue #266, opened 2016-10-28
"Error: unhandled exception: Build failed for package: nimble_test [BuildFailed]" on nimble build


rishavs:
Hi

i created a project using "nimble init". 

In it i added a folder called "src" and one called "bin". 
In the src folder, i have a simple nimfile called main.nim which contains 
`echo "test"`
in it. 
i added `srcDir = "src"`
`binDir = "bin"` and `bin = @["main.nim"]` entries in the nimble file. 
i was expecting that using "nimble build" will compile the main.nim in src into main.exe in bin,
but i get 
`Error: unhandled exception: Build failed for package: nimble_test [BuildFailed]`
instead

My project is here
https://github.com/rishavs/nimble_test/

the error stack is:

```
PS C:\Users\rsharan\Documents\GitHub\nimble_test> nimble build
Looking for sdl2 (any version)...
Dependency already satisfied.
Building nimble_test/main.exe using c backend...
Traceback (most recent call last)
nimble.nim(1004)         nimble
nimble.nim(949)          doAction
nimble.nim(605)          build
nimble.nim(323)          buildFromDir
Error: unhandled exception: Build failed for package: nimble_test [BuildFailed]
```

I am on windows 10 anniversary edition.
Nim is:

```
PS C:\Users\rsharan\Documents\GitHub\nimble_test> nim --version
Nim Compiler Version 0.15.2 (2016-10-23) [Windows: amd64]
Copyright (c) 2006-2016 by Andreas Rumpf

git hash: 5cb50364eb1f70677f619882cfafbd3ede45356f
active boot switches: -d:release
```

dom96:
Works for me on OS X.

I assume compiling the main.nim file manually works, right?

What version of Nimble do you have?

rishavs:
Hi @dom96 

I am assuming this bug is only in windows 10. On Glitter other users on MacOS and Linux didnt get it either.

My nimble version is 

```
PS C:\WINDOWS\system32> nimble -v
nimble v0.7.10 compiled at 2016-10-23 10:31:22
```

crohlfs:
Hi there, I'm also getting this error on Windows 10. After adding extra logging (echoing the error) I got

```'C:\Program' not in PATH.```

After making this change to add quotes around nimBin https://github.com/crohlfs/nimble/commit/4fa4404df831f10e072e872580bca0a57f0a83b5 it seems to be fixed. Not sure if this change would have side effects on other systems or not.

[...]

PR #279, opened 2016-11-21
Fix issues with spaces in bin path

crohlfs:
'build' and 'compile' commands would fail on windows when the nim binary path contained spaces.

Hopefully this doesn't any other systems

As per https://github.com/nim-lang/nimble/issues/266


dom96:
Thank you guys! This is great :)

PR #282, opened 2016-12-10
Fixed installation from git dir

yglukhov:


dom96:
This feel very hackish :)

yglukhov:
Maybe, but it relies on documented nim features =). Would look nicer if thisModulePath was defined somewhere in system.nim

PR #283, opened 2016-12-22
outline of how the support for native package managers should look like

Araq:
To start discussions.

The global 'foreignDeps' needs to be used by Nimble afterwards to either produce a long 'echo' statement what to do to finish the installation or to run these commands directly.


src/nimble.nim
```diff
@@ -1028,6 +1028,10 @@ proc doAction(options: Options) =
     let (_, pkgInfo) = install(options.action.packages, options)
     if options.action.packages.len == 0:
       nimScriptHint(pkgInfo)
+    if pkgInfo.foreignDeps.len > 0:
+      echo("To finish the installation, run: ")
```
dom96: ``echo`` should no longer be used in Nimble to display information. Use the new ``display`` proc:

```nim
display("Hint:", "To finish the installation, run:", Warning, HighPriority)
```

src/nimble.nim
```diff
@@ -1028,6 +1028,10 @@ proc doAction(options: Options) =
...
+      for i in 0..<pkgInfo.foreignDeps.len:
+        echo(pkgInfo.foreignDeps[i])
```
dom96: This should also use ``display``, but the text should be indented :)

src/nimblepkg/nimscriptsupport.nim
```diff
@@ -200,13 +200,21 @@ proc findNimscriptApi(options: Options): string =
 
 proc getNimPrefixDir(): string = splitPath(findExe("nim")).head.parentDir
 
+when declared(ModuleGraph):
```
dom96: What is the purpose of this?

dom96: and why is it a global variable?

Araq: global because we need to share the state here. 'declared' to make nimble compile with older versions of the compiler's API.

PR #295, opened 2016-12-29
Fix git ls-remote parsing

### LINKED ISSUE #294: Warnings by git ls-remote lead to crashes
The code in `download.nim` assumes that `git ls-remote` only outputs references. This isn't the case though; for example, a warning like this can also be displayed:
```
Warning: Permanently added 'github.com,12.34.56.78' (RSA) to the list of known hosts.
```
This gets parsed as a tag, and then (if no other tags are found) injected into a command line without escaping. This breaks package installation, and, consequentially, my CI script ^_^

nigredo-tori:
Fixes #294


dom96:
Nice and easy fix. I love it. Thank you!

PR #296, opened 2016-12-30
Change the way `dump` locates the project file

### LINKED ISSUE #291: dump doesn't use the provided package name
Although help message shows `nimble dump` to accept an optional package name, this name is not used - only information about the package in the current directory is displayed. In my opinion, it would be better to dump the information about the specified package instead.

nigredo-tori:
- Look in current directory (only if no argument is present).
- If the argument is a path to a project file - use it.
- If the argument is a path to a directory - try to find a project file there.
- Try to interpret the argument as the name of an installed package.

Fixes #291


dom96:
This is beautiful, thank you!

PR #297, opened 2016-12-30
Remove quotes around "test"

luked99:
The example for adding a test section uses quotes around the
test target, "test".

However, this doesn't work, as it's not a valid identifier and
nimble package builds fail with:

    Error: identifier expected, but found '`"test" Task`'.

Remove the quotes in the example.


dom96:
Thanks!

luked99:
The test has failed, seems a bit surprising!

   Error: execution of an external program failed: './tester '

dom96:
yeah, an output buffering issue I think.

PR #325, opened 2017-01-28
whooops, missed part two of #324

samdmarshall:
This is part two of the changes necessary for #324 to be successful, i somehow forgot to commit this to my PR yesterday >.<

### DIFF
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -167,7 +167,7 @@ proc publish*(p: PackageInfo) =
   var url = ""
   var downloadMethod = ""
   if dirExists(os.getCurrentDir() / ".git"):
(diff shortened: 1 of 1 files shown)


dom96:
Why is there two different places that this URL is specified? Can the duplication be removed?

samdmarshall:
It looks like the first one i changed is used later and specifically composes it with username and password, whereas the one in this review is just needing the raw url of the repo. The first changeset also silently failed as it was trying to do `git push <url>` but ended up only doing `git push` which happens to also work??

Seems like a lot of this publish code needs to be hardened against error cases.

dom96:
I'll merge it anyway, if you've got a second then please create another PR to refactor the code a bit :)

PR #337, opened 2017-03-11
Access GitHub API token through environment variable

samdmarshall:
This adds functionality to default to using an environment variable for getting the user's GitHub API Token through an environment variable as a higher preference than one written to disk. It also adds an additional line of displayed text to inform the user that their access token will be written to disk (and includes the location of that file). 

While i think the ability to reuse the access token (rather than request a new one each time) is great, the implementation caught me off guard with leaking this to disk without informing the user of it happening. This changeset preserves the existing behavior but uses it as a fall-back to a more standard approach to providing access to api tokens/secrets.

### DIFF
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -51,6 +51,9 @@ proc getGithubAuth(cfg: Config): Auth =
   # always prefer the environment variable to asking for a new one
   if existsEnv(ApiTokenEnvironmentVariable):
     result.token = getEnv(ApiTokenEnvironmentVariable)
+    display("Info:", "Using the '" & ApiTokenEnvironmentVariable & 
+            "' environment varaible for the GitHub API Token.",
+            priority = HighPriority)
   else:
     # try to read from disk, if it cannot be found write a new one
     try:
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -16,6 +16,7 @@ type
 
 const
   ApiKeyFile = "github_api_token"
+  ApiTokenEnvironmentVariable = "NIMBLE_GITHUB_API_TOKEN"
 
 proc userAborted() =
   raise newException(NimbleError, "User aborted the process.")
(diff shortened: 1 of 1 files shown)


src/nimblepkg/publish.nim
```diff
@@ -25,23 +26,37 @@ proc createHeaders(a: Auth): string =
...
-    sleep(5000)
-    display("Info:", "Your default browser should open with the following URL: " &
-            "https://github.com/settings/tokens/new", priority = HighPriority)
-    sleep(3000)
-    openDefaultBrowser("https://github.com/settings/tokens/new")
-    result.token = promptCustom("Personal access token?", "").strip()
-    writeFile(cfg.nimbleDir / ApiKeyFile, result.token)
+  # always prefer the environment variable to asking for a new one
+  if existsEnv(ApiTokenEnvironmentVariable):
+    result.token = getEnv(ApiTokenEnvironmentVariable)
```
dom96: Please add a high priority ``display`` message stating that the env variable was used for the api token.

src/nimblepkg/publish.nim
```diff
@@ -25,23 +26,37 @@ proc createHeaders(a: Auth): string =
...
+  else:
+    # try to read from disk, if it cannot be found write a new one
+    try:
+      result.token = readFile(cfg.nimbleDir / ApiKeyFile)
```
dom96: Also add a message here stating that the API token was read from a file.

src/nimblepkg/publish.nim
samdmarshall: ok this should be updated now.

dom96: You made a typo and also didn't add a message for a file being used.

samdmarshall: oops, sorry. both of those things should be corrected now.

src/nimblepkg/publish.nim
```diff
@@ -25,23 +26,43 @@ proc createHeaders(a: Auth): string =
...
+  display("Info:", "Your default browser should open with the following URL: " &
+          "https://github.com/settings/tokens/new", priority = HighPriority)
+  sleep(3000)
+  openDefaultBrowser("https://github.com/settings/tokens/new")
+  let token = promptCustom("Personal access token?", "").strip()
+  # inform the user that their token will be written to disk
+  let token_write_path = cfg.nimbleDir / ApiKeyFile
+  display("Info:", "Writing access token to file:" & token_write_path, 
+          priority = HighPriority)
+  writeFile(token_write_path, token)
```
dom96: Sorry. I just noticed this. Please use camelCase.

src/nimblepkg/publish.nim
```diff
@@ -25,23 +26,43 @@ proc createHeaders(a: Auth): string =
...
+  # always prefer the environment variable to asking for a new one
+  if existsEnv(ApiTokenEnvironmentVariable):
+    result.token = getEnv(ApiTokenEnvironmentVariable)
+    display("Info:", "Using the '" & ApiTokenEnvironmentVariable & 
+            "' environment variable for the GitHub API Token.",
+            priority = HighPriority)
+  else:
+    # try to read from disk, if it cannot be found write a new one
+    try:
+      let api_token_file_path = cfg.nimbleDir / ApiKeyFile
```
dom96: Use camelCase here too.

dom96:
Thanks and sorry for the back and forth :)

samdmarshall:
no worries! thanks for the feedback!

Issue #339, opened 2017-03-31
Nimble does not properly handle flags on the command line


andreaferretti:
Example: `nimble c -d:release server.nim`:

```
     Error: Traceback (most recent call last)
        ... nimble.nim(1079)         nimble
        ... options.nim(313)         parseCmdLine
        ... options.nim(279)         parseFlag
        ... Unknown option: --d

```

PR #342, opened 2017-04-07
Fixed https://github.com/nim-lang/nimble/issues/339

andreaferretti:
I had to dispatch by action first, in order to handle both the case where `-d` is an alias for `depsonly` and the case where `-d` should be passed to the nim compiler. In general it makes more sense to handle aciton specific flags after having found which action we are in


dom96:
> In general it makes more sense to handle aciton specific flags after having found which action we are in

Indeed. I was planning on refactoring it this way. Thanks :)

Issue #123, opened 2015-05-18
description field missing of some packages


pyloor:
Hi,
currently I got the following error message:

> FAILURE: Package in packages.json file does not contain a description field.

A short grepping shows that the desc field seems mssing on 7 packages

System:
Debian Stretch amd64
nimble v0.6.0 compiled at 2015-05-18 02:55:14
Nim Compiler Version 0.11.2 (2015-05-18) [Linux: amd64]

Iancurtis:
me too . get my foot wet in nim just a hour ago.

Iancurtis:
I guess there's something wrong with json parser.

dom96:
Should be fixed now.

PR #343, opened 2017-04-07
Fixed https://github.com/nim-lang/nimble/issues/305

### LINKED ISSUE #305: Show raw output
Nimble changes the output color and formatting when executing tasks. Some tools, such as `unittest` or  the compiler itself, use color to display information in a more readable way.

It would be nice if there was an option to show raw output (without buffering it, adding indentation or changing color). This may also be useful in order to write tasks whose output needs to be predictable - maybe because it is piped in other commands.

andreaferretti:
It seems to me that this should fix #305, which is the main obstacle I find in using nimble as a task runner

### DIFF
--- a/tests/tester.nim
+++ b/tests/tester.nim
@@ -188,9 +188,9 @@ test "can use nimscript's setCommand":
 test "can use nimscript's setCommand with flags":
   cd "nimscript":
     let (output, exitCode) = execNimble("--debug", "cr")
-    # let lines = output.strip.splitLines()
+    let lines = output.strip.splitLines()
     check exitCode == QuitSuccess
-    # check "Hello World".normalize in lines[^2].normalize
+    check inLines(lines, "Hello World")
 
 test "can use nimscript with repeated flags (issue #329)":
   cd "nimscript":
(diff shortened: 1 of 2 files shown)


tests/tester.nim
```diff
@@ -188,9 +188,9 @@ test "can use nimscript's setCommand":
 test "can use nimscript's setCommand with flags":
   cd "nimscript":
     let (output, exitCode) = execNimble("--debug", "cr")
-    let lines = output.strip.splitLines()
+    # let lines = output.strip.splitLines()
```
dom96: Why was this commented out?

andreaferretti:
Because now the output is live, hence it is not captured in your test :-)

dom96:
huh? Why not? Why does it being live prevent it being captured?

andreaferretti:
Umm, I will check better and figure out, on a second thought I think you
are right

dom96:
I think the order of the messages is simply different. All you need to do is use the ``inLines`` procedure defined at the top of the tester (other tests use it too, and probably all of them should use it).

dom96:
Thanks! btw you don't need to write "fixes nim-lang#123" in the commit message, "fixes #123" is enough.

andreaferretti:
You are right, it was just a matter of using `inLines` :-)

Issue #338, opened 2017-03-24
Nimble throws exception based on srcDir contents 


samdmarshall:
Having the `srcDir` path defined in the nimble package end with a trailing `/` causes nimble to throw an exception. See difference between lines 11 and 47 here: https://gist.github.com/samdmarshall/778c2fdf7458769d8ac85608ca632481

PR #345, opened 2017-04-22
Modifying how package install path is validated

samdmarshall:
Fix:
  * This addresses a bug where nimble will throw an exception in the middle of installation due to trying to evaluate paths that are not normalized. This was fixed by adding some additional validation checks that involve comparing paths by normalizing them first. [#338]

### DIFF
--- a/src/nimblepkg/tools.nim
+++ b/src/nimblepkg/tools.nim
@@ -79,7 +79,14 @@ proc changeRoot*(origRoot, newRoot, path: string): string =
   ## newRoot:  /home/test/
   ## path:     /home/dom/bar/blah/2/foo.txt
   ## Return value -> /home/test/bar/blah/2/foo.txt
-  if path.startsWith(origRoot):
+
+  ## The additional check of `path.samePaths(origRoot)` is necessary to prevent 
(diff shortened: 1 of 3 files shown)


samdmarshall:
I'd be happy to, i'm not sure how you would like me to test this. Would modifying an existing test to show the regression be sufficient or should i be making a new test case to have it try against?

dom96:
A new test case would be great. I think what you need to do is the following:

* Create a new ``issue338`` directory inside the ``tests`` directory.
* Create a package inside that directory that reproduces issue #338.
* Add code to the ``tester`` to test the installation of that package.

Be sure to test that the test fails when your patch is not applied, and that it works when it is.

samdmarshall:
Ace, thanks for the info. I think I should be able to get to doing this later today.

samdmarshall:
@dom96 I had some trouble with the rest of the test suite (inconsistent results of the tests between runs) but I was able to verify that the new test that I added for this was able to fail without the code-fix applied and didn't fail when it was applied.

dom96:
Brilliant, thanks!

Issue #334, opened 2017-02-13
[Feature request] Allow passing flags to `nimble build`


moigagoo:
It would be really useful to be able to pass flags to the compiler with `nimble build`.

For example, when I want to debug the package, I'd launch `nimble build --debugger:native` while keeping `config.nims` clean and suitable for release, for example: `-d:ssl`.

dom96:
Should be easy to implement this in case anyone wants to take a crack at it :)

dom96:
Fixed via #347.

weskerfoot:
Where is the documentation for this feature?

PR #347, opened 2017-04-27
Change to pass extra parameters to compiler

antizealot1337:
This PR changes `nimble build` to pass along extra command line arguments to the compiler as mentioned in #334. Please let me know if there's anything I can change to make it better.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -285,15 +285,15 @@ proc processDeps(pkginfo: PackageInfo, options: Options): seq[string] =
   for i in reverseDeps:
     addRevDep(options, i, pkginfo)
 
-proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], forRelease: bool) =
+proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], args: var string) =
   ## Builds a package as specified by ``pkgInfo``.
   if pkgInfo.bin.len == 0:
     raise newException(NimbleError,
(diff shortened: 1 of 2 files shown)


Araq:
Didn't review the implementation, but this feature is **important**.

antizealot1337:
The second commit is labelled wrong. It should read "Removed excess releaseOpt variable since args are passed in" or something similar. I've never made a pull request before so I'm not sure how to fix it. Would `git commit --amend && git push -f` screw it up?

src/nimble.nim
```diff
@@ -285,16 +285,14 @@ proc processDeps(pkginfo: PackageInfo, options: Options): seq[string] =
   for i in reverseDeps:
     addRevDep(options, i, pkginfo)
 
-proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], forRelease: bool) =
+proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string], args: var string) =
```
dom96: I would prefer if this took ``seq[string]``

dom96:
But i'll merge anyway.

dom96:
Thanks for the PR!

antizealot1337:
Thanks for the merge :). I agree that the `var string` should be `var seq[string]`. At first I thought the method would be fine with a `string` and didn't change it when it was apparent it wouldn't work like that. I'll send a pull request that changes to `seq[string]`. I think it is a superior implementation.

PR #348, opened 2017-05-05
Change buildFromDir to accept a seq[string] instead of a string.

### LINKED ISSUE #334: [Feature request] Allow passing flags to `nimble build`

antizealot1337:
Follow up to #347.


dom96:
Don't know why it took me so long to accept this. I hope I didn't forget something important :)

PR #360, opened 2017-06-01
Move the .nimble folder away from the home root.

xyz32:
There are way to many .something folders in the home root.


dom96:
All package managers place it in ``~``, if you dislike that location then you can change it via the config file: https://github.com/nim-lang/nimble#configuration

PR #364, opened 2017-06-08
nimble accepts a .nimble file that is a link

jxy:


dom96:
Hrm, what use case does this handle?

jxy:
1. Out-of-tree build.
2. Build with shared and/or immutable source directory.
3. Build with different local configurations.

One example: https://github.com/jcosborn/qex/blob/devel/qex.nimble

Besides, I usually use tools that do not differentiate files with links.

dom96:
Fair enough. :)

PR #371, opened 2017-06-25
Allow specifying local json package list files

### LINKED ISSUE #368: Accept local packages.json files
Currently nimble can only be pointed at packages.json accessible via a URL. We should allow users to install json files locally.

Currently this is possible via a small hack (manually placing a `packages_foo.json` file in ~/.nimble), so it's mostly a matter of exposing it as a public option.

subsetpark:
Fixes #368 .


dom96:
Beautiful. Tests and everything is awesome :D

PR #378, opened 2017-07-19
Add test target

Nycto:
As you can likely guess, this adds `nimble test` as a target. This target iterates through all the nim files in the `tests` directory and executes them. If any fail, the target fails.

My only concern with this change is that there are quite a few packages that have defined a custom `test` target of their own. In fact, `nimble` itself is one of them. By defining _this_ test target, any custom defined targets will cease running. However, a quick scan through Github shows that quite a bit are doing exactly what this change implements.

One option I considered to assuage this behavior is by allowing custom targets to override default targets. But this is a bigger change that I thought best discussed before being implemented.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -673,6 +673,28 @@ proc execBackend(options: Options) =
         [backend, args, bin], showOutput = true)
   display("Success:", "Execution finished", Success, HighPriority)
 
+proc tempOutArg(file: string): string =
+  ## Returns the `--out` argument to pass to the compiler, using a temp file
+  let (_, name, _) = splitFile(file)
+  let dir = getNimbleTempDir() / "tests"
+  createDir(dir)
+  result = "--out:" & (dir / name)
+
+proc test(options: Options) =
+  ## Executes all tests
+  var files = toSeq(walkDir(getCurrentDir() / "tests"))
+  files.sort do (a, b: auto) -> int:
+    result = cmp(a.path, b.path)
+
+  for file in files:
+    if file.path.endsWith(".nim") and file.kind in { pcFile, pcLinkToFile }:
+      var optsCopy = options
+      optsCopy.action.file = file.path
+      optsCopy.action.backend = "c -r"
+      optsCopy.action.compileOptions.add("--path:.")
+      optsCopy.action.compileOptions.add(file.path.tempOutArg)
+      execBackend(optsCopy)
+
 proc search(options: Options) =
   ## Searches for matches in ``options.action.search``.
   ##
@@ -1030,6 +1052,8 @@ proc doAction(options: Options) =
     listPaths(options)
   of actionBuild:
     build(options)
+  of actionTest:
+    test(options)
   of actionCompile, actionDoc:
     execBackend(options)
   of actionInit:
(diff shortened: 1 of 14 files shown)


dom96:
This is a beautiful PR. Even comes with tests 👏 

Sorry it took me so long to get to it. In general I love it but it would be great if we could still support people's custom ``test`` targets. Ideally if a ``test`` target doesn't exist in the nimble file then we can fall back to the one you implemented. Which now that I read more I see that you've suggested.

Do you want to implement that? I might do it myself if I don't hear back from you (I'm likely going to release a new Nimble this weekend).

src/nimble.nim
```diff
@@ -673,6 +673,28 @@ proc execBackend(options: Options) =
         [backend, args, bin], showOutput = true)
   display("Success:", "Execution finished", Success, HighPriority)
 
+proc tempOutArg(file: string): string =
```
dom96: Is this so that binaries are not created inside the user's ``tests`` directory? I wonder if it would be useful to not do that.

src/nimble.nim
```diff
@@ -673,6 +673,28 @@ proc execBackend(options: Options) =
...
+  ## Returns the `--out` argument to pass to the compiler, using a temp file
+  let (_, name, _) = splitFile(file)
+  let dir = getNimbleTempDir() / "tests"
+  createDir(dir)
+  result = "--out:" & (dir / name)
+
+proc test(options: Options) =
+  ## Executes all tests
+  var files = toSeq(walkDir(getCurrentDir() / "tests"))
+  files.sort do (a, b: auto) -> int:
```
dom96: Prefer `=>` from the `future` module when possible.

src/nimble.nim
```diff
@@ -673,6 +673,28 @@ proc execBackend(options: Options) =
...
+    result = cmp(a.path, b.path)
+
+  for file in files:
+    if file.path.endsWith(".nim") and file.kind in { pcFile, pcLinkToFile }:
+      var optsCopy = options
+      optsCopy.action.file = file.path
+      optsCopy.action.backend = "c -r"
```
dom96: the ``-r`` isn't a backend, it belongs in the ``compileOptions``.

src/nimble.nim
```diff
@@ -673,6 +673,28 @@ proc execBackend(options: Options) =
...
+      optsCopy.action.compileOptions.add("--path:.")
+      optsCopy.action.compileOptions.add(file.path.tempOutArg)
+      execBackend(optsCopy)
+
```
dom96: I'm not sure whether ``execBackend`` handles this but we need to be sure to exit with an error exit code.

Also, test fail/pass messages would be nice.

tests/tester.nim
```diff
@@ -33,6 +33,7 @@ proc execNimble(args: varargs[string]): tuple[output: string, exitCode: int] =
   quotedArgs = quoted_args.map((x: string) => ("\"" & x & "\""))
 
   result = execCmdEx(quotedArgs.join(" "))
+  checkpoint(result.output)
```
dom96: 👍

Nycto:
Thanks for taking a look!

I probably won't have time to make changes until mid-week, so if you've got the time and desire, that's totally cool.

Nycto:
Thanks for following through on this one! Much appreciated. My apologies for not having a chance to fix it up, but I'm really happy to see this feature land.

dom96:
No worries. Thanks for the PR :D

PR #383, opened 2017-07-26
Added better url detection to publish

waylon531:
Publish will now look at the hostname part of the ssh url to figure out
what git server to use. This definitely works with both github and
gitlab, and adds gitlab support to `nimble publish`.

### DIFF
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -206,7 +206,7 @@ proc publish*(p: PackageInfo, o: Options) =
     if parsed.scheme == "":
       # Assuming that we got an ssh write/read URL.
       let sshUrl = parseUri("ssh://" & url)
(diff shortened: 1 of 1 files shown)


Araq:
Awesome! @dom96 please have a look.

src/nimblepkg/publish.nim
```diff
@@ -206,7 +206,7 @@ proc publish*(p: PackageInfo, o: Options) =
     if parsed.scheme == "":
       # Assuming that we got an ssh write/read URL.
       let sshUrl = parseUri("ssh://" & url)
-      url = "https://github.com/" & sshUrl.port & sshUrl.path
+      url = "https://" & sshUrl.hostname & "/" & sshUrl.port & sshUrl.path
```
dom96: Nice catch.

I'm pretty sure this URL construction will still fail if a port is specified though. But i'll merge anyway.

PR #387, opened 2017-08-10
Removed deprecated warnings, also tiny refactoring

TiberiumN:
Added "http" field to "Auth" type in publish.nim to remove deprecated getContent warnings, headers are now added with a normal way, changed old % json macro to %*
Also removed "when isMainModule" in publish.nim, since code in it is broken


dom96:
Looks good, thanks!

PR #401, opened 2017-09-03
Fixed #400

### LINKED ISSUE #400: Mixed (lib/bin) package: binary fails to see lib files
Package structure:
```
- foo.nimble
- foo
  |- foo1.nim
  |- foo2.nim
```
foo.nimble:
```nim
version       = "0.1.0"
author        = "Anonymous"
description   = "foo"
license       = "MIT"
bin = @["foo/foo1"]
```
foo1.nim:
```nim
import foo.foo2
```
```
nimble install -y
  Verifying dependencies for foo@0.1.0
   Warning: No nimblemeta.json file found in /Users/yglukhov/.nimble/pkgs/nimble-#head
 Installing foo@0.1.0
   Building foo/foo/foo1 using c backend
     Error: Build failed for package: foo
        ... Details:
        ... Execution failed with exit code 1
        ... Command: "/Users/yglukhov/Projects/Nim/bin/nim" c --noBabelPath -d:release -o:"/Users/yglukhov/Projects/foo/foo/foo1" "/Users/yglukhov/Projects/foo/foo/foo1.nim"
        ... Output: Hint: used config file '/Users/yglukhov/Projects/nim/config/nim.cfg' [Conf]
        ... Hint: system [Processing]
        ... Hint: foo1 [Processing]
        ... foo1.nim(1, 11) Error: cannot open 'foo/foo2'
```

yglukhov:


dom96:
Thank you for this quick fix.

However, I'm not sure this is a bug. Considering the case I ran into recently: https://github.com/nim-lang/Nim/issues/6315

If you attempt to install this package, then compile the local copy you will run into the same issue. I think what you need to do is `import foo2`

yglukhov:
@dom96, ok, i thought about this a bit more, i even tried to use relative imports, but turns out that doesn't really work in my case. So here it is. I've got a pretty significant codebase of a lib package. The modules import each other with absolute imports. Also i've got a bin in this same package, which doesn't import any of lib files. Until now. I discovered myself in a situation that I can't import any of lib modules to the bin modules because I'll have to modify a whole lot of my files to use relative imports.

So IMO we should definitely work towards allowing absolute imports.

yglukhov:
@dom96, btw, I can't reproduce your bug with latest nimble + this PR + latest nim.

dom96:
Okay, so in order to reproduce my bug you have to add `import foo2` to foo1.nim. Compiling `foo1.nim` will then result in:

```
foo2.nim(1, 2) Error: module names need to be unique per Nimble package; module clashes with /Users/dom/.nimble/pkgs/foo-0.1.0/foo/foo2.nim
```

**Note:** The `foo` package needs to be installed for this to happen (I had to comment out the `bin = ...` line to get it to install).

This is really subtle, and you can get into some really confusing situations:

* Create a `foo3.nim`
* Add `import foo3` into the foo1.nim file (and forget about my earlier addition, just have `import foo.foo2, foo3` in the foo1.nim file)
* Install the package.
* `nim c foo/foo1`

The binary will use `foo2` from `/Users/dom/.nimble/pkgs/foo-0.1.0/foo/foo2.nim` and `foo3` from `./foo/foo3`.

We should really make sure to avoid such situations. I hope this makes sense, let me know your thoughts.

yglukhov:
@dom96, I'm back to work on this! =)
And again, I can't reproduce this with my pr.
```
$ cd tests/issue400
$ nimble install -y
  Verifying dependencies for foo@0.1.0
 Installing foo@0.1.0
   Building foo/foo/foo1 using c backend
   Success: foo installed successfully.
$ nimble install -y # again
  Verifying dependencies for foo@0.1.0
 Installing foo@0.1.0
   Building foo/foo/foo1 using c backend
    Prompt: foo@0.1.0 already exists. Overwrite? -> [forced yes]
   Success: foo installed successfully.
```
The results are the same regardless I `import foo.foo2` or `import foo2` from `foo1.nim`. Also I'm not sure how it can be otherwise, as nimble runs nim with `--noBabelPath` flag so that nim should be unaware of all the installed packages except the exact dependencies provided with `--path` args. Here's the code that does it: https://github.com/nim-lang/nimble/pull/401/files#diff-fe0eee2675411a16f5bb411dce90d5b1R241. Note the `--noBabelPath` option.

yglukhov:
@dom96, how about you write a test case that breaks my pr, because, honestly, I tried ;)

yglukhov:
Closing this. Likely needs to be done differently.

PR #404, opened 2017-09-08
Fixes #403

### LINKED ISSUE #403: nimble path is inconsistent with nimble develop
For `nimble develop`ed project `nimble path` shows a path to `~/.nimble/pkgs/...`.

yglukhov:


src/nimble.nim
```diff
@@ -615,15 +615,23 @@ proc listPaths(options: Options) =
...
-
-      if hasSpec:
+      var nimbleFile = path / name.addFileExt("nimble")
+      var nimbleLinkTargetPath: string
+      if not nimbleFile.existsFile:
+        let nimbleLinkFile = path / name.addFileExt("nimble-link")
+        if fileExists(nimbleLinkFile):
+          let lns = readFile(nimbleLinkFile).splitLines()
+          nimbleFile = lns[0]
+          nimbleLinkTargetPath = lns[1]
```
dom96: I'm fairly sure there is a function that deals with this logic, please use it instead of duplicating the code.

yglukhov:
> I'm fairly sure there is a function that deals with this logic

Could not find any, so added one.

src/nimble.nim
```diff
@@ -615,15 +615,23 @@ proc listPaths(options: Options) =
       if kind != pcDir or not path.startsWith(options.getPkgsDir / name):
         continue
 
-      let
-        nimbleFile = path / name.addFileExt("nimble")
-        hasSpec = nimbleFile.existsFile
-
-      if hasSpec:
+      var nimbleFile = path / name.addFileExt("nimble")
```
dom96: Why don't you just use the `findNimbleFile` proc here?

yglukhov: Err. Because we still need to know whether it is a link? Specifically, the `nimbleLinkTargetPath` value.

dom96:
This is what I meant btw: https://github.com/nim-lang/nimble/commit/a02a1431e2167bf2fdf59c25d7a4ed4696c6bbc4

yglukhov:
👍

PR #415, opened 2017-10-14
Fix deprecated warning for getBVal

Yardanico:


dom96:
Maybe you could rebase this? I think tests might pass then.

Yardanico:
@dom96 oh, it seems I can't do that because I changed my nickname on GitHub. I'll create new PR

PR #416, opened 2017-10-18
Put "nimblepkg/nimscriptapi.nim" under nimbleDir, fix #363

### LINKED ISSUE #363: nimble put `/tmp/` to search paths and impair other users
This code
```nim
  let tmpNimscriptApiPath = getTempDir() / "nimblepkg" / "nimscriptapi.nim"
  createDir(tmpNimscriptApiPath.splitFile.dir)
  writeFile(tmpNimscriptApiPath, nimscriptApi)
  appendStr(searchPaths, getTempDir())
```
has a few serious problems.

1. It leaves `/tmp/nimblepkg` owned by the current user and the directory remains after `nimble` finishes, which breaks `nimble` for all other users.
2. `/tmp/` is now in `searchPaths`, which is a big security risk.

I wish there is a way to call `mkstemp(3)` with `nimscript`.  For a quick fix, I would propose `s/getTempDir()/"nimblecache"/`.

jxy:
By default, the module will be under
  $HOME/.nimble/nimblecache/nimblepkg/nimscriptapi.nim
and the following directory will be in the searchPaths
  $HOME/.nimble/nimblecache/

### DIFF
--- a/src/nimblepkg/nimscriptsupport.nim
+++ b/src/nimblepkg/nimscriptsupport.nim
@@ -224,11 +224,12 @@ proc execScript(scriptName: string, flags: Flags, options: Options): PSym =
   let pkgName = scriptName.splitFile.name
 
   # Ensure that "nimblepkg/nimscriptapi" is in the PATH.
-  # TODO: put this in a more isolated directory.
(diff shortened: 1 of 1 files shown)


jxy:
I messed up the history a bit.  If you want a cleaner commit history, I could try another push request.

dom96:
Looks good, thanks!

Issue #340, opened 2017-04-04
Add NIMBLE_DIR environment variable


moigagoo:
Currently, Nimble installs packages to ~/.nimble, a hardcoded path. You can change it for a particular package with nimbleDir, but there is no way to set it once and for all.

I think Nimble should respect `NIMBLE_DIR` environment variable if it's set. This achieves the following goals:
- anyone can tune it to their liking (https://github.com/nim-lang/nimble/issues/141)
- you can easily change the default Nimble path in future versions by simply modifying the variable (for example to ~/nimble, as proposed in #141)
- scoop will be able to make binaries installed with Nimble available in the console (currently, it doesn't allow adding directories outside Nim directory to PATH)

This change is backward-compatible.

dom96:
This would require support in the Nim compiler. 

Personally I'm not sure it's needed. You can specify this in [Nimble's config file](https://github.com/nim-lang/nimble#configuration) and Nim's config file.

moigagoo:
> This would require support in the Nim compiler.

Hm, I didn't think this required such dramatic modifications.

I guess I better ask the Scoop guy to allow adding arbitrary paths to PATH. You see, to keep installed programs isolated, he allows only subfolders to be added to PATH during installation. So, I could add /path/to/nim/.nimble/bin but not ~/.nimble/bin. Setting NIMBLE_DIR to ~/.nimble/bin would solve the issue for me.

moigagoo:
So, the point of having the environment variable is not only to let the user decide where to install Nimble packages, but also to let the system know where to look for them.

dom96:
Yes, can this be resolved?

moigagoo:
Sorry, do you want to close the issue? If yes, then go ahead. It's your repo after all :-)

dom96:
Were you able to get Nimble working in Scoop without this change?

moigagoo:
Nope. Other Scoop packages I checked don't have this issue, so it's very unlikely that Scoop devs will change the current PATH policy just for Nim. I know I wouldn't have.

To be fair, I haven't reported the feature request. Maybe it's worth a try although I doubt it would be good for Scoop.

pb-cdunn:
@dom96, In order to use `nimble` for enterprise work, we need (I think) either a `NIMBLE_DIR` environment-variable or a `nimble --nimbleDir` command-line flag. (Or maybe `-d:nimbleDir=dir`?)

[...]

PR #417, opened 2017-10-20
Initialize nimbleDir option from the environment variable NIMBLE_DIR

jxy:
See #340

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -786,9 +786,9 @@ proc uninstall(options: Options) =
       if revDeps.len == 1:
         reason = "$1 ($2) depends on it" % [revDeps[0].name, $revDeps[0].ver]
       else:
-        for i in 0 .. <revDeps.len:
+        for i in 0 ..< revDeps.len:
           reason.add("$1 ($2)" % [revDeps[i].name, $revDeps[i].ver])
-          if i != <revDeps.len:
+          if i != revDeps.len-1:
             reason.add ", "
         reason.add " depend on it"
 
@@ -802,7 +802,7 @@ proc uninstall(options: Options) =
       raise newException(NimbleError, "\n  " & errors.join("\n  "))
 
   var pkgNames = ""
-  for i in 0 .. <pkgsToDelete.len:
+  for i in 0 ..< pkgsToDelete.len:
     if i != 0: pkgNames.add ", "
     let pkg = pkgsToDelete[i]
     pkgNames.add("$1 ($2)" % [pkg.name, pkg.specialVersion])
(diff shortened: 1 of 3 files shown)


jxy:
Any comment about this?  Did I miss anything?

src/nimblepkg/options.nim
```diff
@@ -305,7 +305,7 @@ proc parseFlag*(flag, val: string, result: var Options, kind = cmdLongOption) =
 proc initOptions*(): Options =
   result.action.typ = actionNil
   result.pkgInfoCache = newTable[string, PackageInfo]()
-  result.nimbleDir = ""
+  result.nimbleDir = getEnv("NIMBLE_DIR")
```
dom96: Please output a warning or a hint message that this environment variable is being used.

jxy: I can do that.  I changed it here so that the command line option parsed later can override this environment variable, and when the command line option takes precedence, I guess we need another hint to disambiguate nimble's decision.  Do you think it's okay to do that, or would it be too chatty?

dom96: It will be best if a message is only printed when the environment variable is used. So maybe instead of setting it here you could check whether ``nimbleDir`` is empty later down the line?

src/nimble.nim
```diff
@@ -498,12 +498,15 @@ proc build(options: Options) =
...
-  let bin = options.action.file
+  let
+    bin = options.action.file
+    binDotNim = bin.addFileExt("nim")
   if bin == "":
     raise newException(NimbleError, "You need to specify a file.")
 
-  if not fileExists(bin):
-    raise newException(NimbleError, "Specified file does not exist.")
+  if not (fileExists(bin) or fileExists(binDotNim)):
```
dom96: Nice, I see you've fixed another issue :)

src/nimblepkg/options.nim
```diff
@@ -352,6 +352,11 @@ proc parseCmdLine*(): Options =
   if result.action.typ == actionNil and not result.showVersion:
     result.showHelp = true
 
+  # Inform user that we use their environment variables.
+  if result.getNimbleDir == getEnv("NIMBLE_DIR"):
+    display("Info:", "Using the 'NIMBLE_DIR' environment variable.",
```
dom96: This will show the message even if ``NIMBLE_DIR`` is empty which isn't what we want IMO. Sorry, about the back and forth. I can fix this myself if you'd like.

jxy: Really?  ``result.getNimbleDir`` will not return an empty string.

dom96: oh, you're right. My bad.

jxy: It's a bit chatty when you use it.  Perhaps we should implement something like ``go env``.

dom96: oh, then you should make it a warning. IIRC Nimble keeps track of which warnings have been shown and ensures they are only shown once.

jxy: alright, check #450

PR #427, opened 2017-11-12
Enhance nimble as a library

bluenote10:
It would be really nice if the high level functionality of Nimble would be available to write scripts on top of Nimble. Currently the main `nimble` module doesn't really export anything (except for one proc I think), and this the operations like `init`/`install`/`build` are not available on library level. This PR would allow to e.g. `import nimble` and perform these high level Nimble operations from within Nim.

I have also exported the package validation (stand-alone) function, because I had a use case where running this explicitly was also necessary.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -18,7 +18,7 @@ import nimblepkg/packageinfo, nimblepkg/version, nimblepkg/tools,
 
 import nimblepkg/nimscriptsupport
 
-proc refresh(options: Options) =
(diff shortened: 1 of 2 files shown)


dom96:
At that point I think you're better off simply calling ``nimble init|refresh|...`` instead of importing Nimble (it's a big dependency).

dom96:
Yeah, I still think it's better to just execute nimble.

Issue #428, opened 2017-11-12
Problems with --nimbleDir for certain directory names (starting with nimble?)


bluenote10:
I think there are some issues depending on the name of the `--nimbleDir` passed to Nimble. Given the following dummy .nimble file:

```
# Package

version       = "0.1.0"
author        = "Author"
description   = "dummy"
license       = "MIT"

# Dependencies

requires "nim >= 0.17.3"
```

Trying to run `nimble --nimbleDir=./nimbleDir install` gives the following traceback:

```
   Warning: Package 'dummy' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'dummy', but file 'nimscriptapi.nim' is in a directory named 'nimbleDir/nimblecache/nimblepkg' instead. This will be an error in the future.
      Hint: If 'nimbleDir/nimblecache/nimblepkg' contains source files for building 'dummy', rename it to 'dummy'. Otherwise, prevent its installation by adding `skipDirs = @["nimbleDir/nimblecache/nimblepkg"]` to the .nimble file.
  Verifying dependencies for dummy@0.1.0
 Installing dummy@0.1.0
Traceback (most recent call last)
nimble.nim(1024)         nimble
nimble.nim(963)          doAction
nimble.nim(463)          install
nimble.nim(383)          installFromDir
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
packageinfo.nim(516)     iterInstallFiles
[...]

PR #429, opened 2017-11-12
Terminate copy iteration when hitting in-place nimbleDir

bluenote10:
I think the problem I observed in #428 has nothing to do with the directory name itself. The behavior was just a coincidence of the order `walkFiles` works. The real issue is probably that the iteration does not terminate when the `nimbleDir` is part of the directory to be installed.


bluenote10:
Hm, the test works for me locally -- will have to investigate what's different on Travis.

PR #435, opened 2017-11-30
Make nimble init create a package without prompting unless --prompt is provided

antizealot1337:
My attempt at implementing #288. Let me know if there's anything you'd like changed.

Issue #288, opened 2016-12-25
nimble init shouldn't prompt by default


dom96:
The ``init`` command should analyse the project and create an appropriate ``.nimble`` file with no prompts. The current behaviour should be behind a ``--prompt`` flag.

antizealot1337:
I think I'd like to add this but don't know which way to go. Personally, I like the not prompting by default but figure it would be very unexpected for most users to get a new nimble version and `init` act very different.

I also have a question of the flag to change the behavior. I figure if not prompting `--prompt` would be the flag name. I'm not a big fan of `--no-prompt` for not prompting so how do you feel about `--guess`, `--auto`, or maybe even `--default`/`--defaults`? Another flag alternative could be `--prompt:on|off` which resembles how the compiler enables and disables checks. It may allow changing the default behavior of `init` in the future without much friction.

I'm also unclear how much analysis you would want nimble to do for an existing project?

dom96:
> Personally, I like the not prompting by default but figure it would be very unexpected for most users to get a new nimble version and init act very different.

That's true, I think showing a "Hint: New version of Nimble no longer prompts for `init`. See <url> for more details" message would help the expectations.

I'm still leaning towards Nimble not prompting by default, so ``--prompt`` sounds good.  You could probably support the trifecta: ``--prompt``, ``--prompt:off``, ``--prompt:on`` :)

> I'm also unclear how much analysis you would want nimble to do for an existing project?

For a first version it's important to support the common case: you've just created a new directory called "myawesomeproject" and want to make it into a Nimble package. You can infer the same things that Nimble already infers: the project name (based on directory name) and author name (based on Git info IIRC).

You could also create a nice directory structure for the user:

* src/PkgName/
* src/PkgName.nim
* tests/sampletest.nim

Future versions would add support for existing projects being turned into a Nimble package (but this should be pretty rare)

antizealot1337:
Awesome. Thanks for the advise. :-D

I'll go ahead and start working on this and get a PR as soon as possible.

FedericoCeratto:
Nimble should still ask for the license!
Also #413 would require prompting for some options.
Removing the prompt by default sounds like surprising the user unnecessarily.
Plus, being shown reasonable guesses (e.g. project name, user name) and simply pressing enter takes very little time and people are starting new projects that frequently.

antizealot1337:
[...]

PR #437, opened 2017-12-04
Reuse -y to mean accept defaults for init command for #288

antizealot1337:
An alternative to #435 for #288. `-y` is reused/overloaded to mean "shut up and give me the defaults".

I had to supply a default description that wasn't empty. If not then right after "Success: Created nimble file" an exception is thrown complaining about an empty description.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -760,6 +760,8 @@ proc init(options: Options) =
     $nimDepDef)
   validateVersion(pkgNimDep)
 
+  let pkgTestDir = "tests"
+
   # Create source directory
   os.createDir(pkgSrcDir)
 
@@ -770,8 +772,26 @@ proc init(options: Options) =
   cd pkgSrcDir:
     pkgName.changeFileExt("nim").writeContents "echo \"Hello, World!\"\n"
 
-  display("Success:", "Created initial source file successfully", Success,
+    display("Success:", "Created initial source file successfully", Success,
+      MediumPriority)
+
+  # Create test directory
+  os.createDir(pkgTestDir)
+
+  display("Success:", "Test directory created successfully", Success,
     MediumPriority)
+
+  cd pkgTestDir:
+    "tester.nims".writeContents("""switch("path", "$$projectDir/../$#")""" %
+      [pkgSrcDir])
+
+    display("Success:", "Test config file created successfully", Success,
+      MediumPriority)
+
+    "tester.nim".writeContents("doAssert(1 + 1 == 2)\n")
+
+    display("Success:", "Test file created successfully", Success,
+      MediumPriority)
 
   # Write the nimble file
   nimbleFile.writeContents """# Package
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -714,21 +714,27 @@ proc init(options: Options) =
   display("Using", "$# for new package name" % [pkgName.escape()],
     priority = HighPriority)
 
+  # Ask for package author
+  proc getAuthor(): string =
+    if findExe("git") != "":
+      let (name, exitCode) = doCmdEx("git config --global user.name")
+      if exitCode == QuitSuccess and name.len > 0:
+        result = name.strip()
+        display("Using", "$# for new package author" % [result.escape()],
+          priority = HighPriority)
+    elif findExe("hg") != "":
+      let (name, exitCode) = doCmdEx("hg config ui.username")
+      if exitCode == QuitSuccess and name.len > 0:
+        result = name.strip()
+        display("Using", "$# for new package author" % [result.escape()],
+          priority = HighPriority)
+    else:
+      result = promptCustom(options, "Your name?", "Anonymous")
+  let pkgAuthor = getAuthor()
+
   # Ask for package version.
   let pkgVersion = promptCustom(options, "Initial version of package?", "0.1.0")
   validateVersion(pkgVersion)
-
-  # Ask for package author
-  var defaultAuthor = "Anonymous"
-  if findExe("git") != "":
-    let (name, exitCode) = doCmdEx("git config --global user.name")
-    if exitCode == QuitSuccess and name.len > 0:
-      defaultAuthor = name.strip()
-  elif defaultAuthor == "Anonymous" and findExe("hg") != "":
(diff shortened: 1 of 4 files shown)


src/nimblepkg/options.nim
```diff
@@ -194,6 +194,16 @@ proc prompt*(options: Options, question: string): bool =
   ## forcePrompts has a value different than dontForcePrompt.
   return prompt(options.forcePrompts, question)
 
+proc promptCustom*(options: Options, question, default: string): string =
+  ## Asks an interactive question and returns the result.
+  ##
+  ## The proc will return "default" without asking the user if the global
+  ## forcePrompts is forcePromptYes.
+  result = if options.forcePrompts == forcePromptYes:
```
dom96: Even though the default is returned, you should still show the prompt and emphasise that the default is chosen. The other prompts do this.

antizealot1337: Okay. I'll fix this. I think I can add a `ForcePrompt` value to `cli.promptCustom` with a default of `dontForcePrompt`. That way `promptCustom` has all the functionality and the change doesn't break other uses of `promptCustom`. What do you think?

dom96: Hrm, not entirely certain what you mean, but sounds like it could work. I'll judge it better when I see it, hopefully it's not too much work.

antizealot1337: I probably didn't explain myself well. Sorry. I'm considering making `cli.promptCusom` similar to `cli.prompt` in that it will accept a `ForcePrompt` argument but have a default of `dontForcePrompt` so other existing uses aren't disrupted.

antizealot1337: I made the changes. If you don't like them I can can show the prompts another way.

antizealot1337:
This is fairly feature complete now. Nimble doesn't prompt for the package name, author (but does indicate the value it uses) or source directory. Nimble won't ask for anything when `-y` is passed. Nimble also creates the `src` and `tests` directories with some initial content.

I was able to get a selection list to work haven't added it to this branch yet. I had an issue with `terminal` where no value was sent for the arrow keys under Windows. I have a solution but feel it's incomplete. I put everything needed to make it work in a new module so it can be removed in case `terminal` or something else offers a better version.

In any case, if you want I can push those changes to this branch and you can review them together or I can hold off.

src/nimblepkg/tools.nim
```diff
@@ -98,6 +98,16 @@ proc copyFileD*(fro, to: string): string =
   copyFileWithPermissions(fro, to)
   result = to
 
+proc writeContents*(filename, contents: string) =
```
dom96: You can achieve the same with just a ``try: writeFile(...) except OSError: raise(NimbleError, ... )`` so I don't think this proc is necessary.

src/nimblepkg/cli.nim
```diff
@@ -146,19 +146,43 @@ proc prompt*(forcePrompts: ForcePrompt, question: string): bool =
...
 proc promptCustom*(question, default: string): string =
-  if default == "":
-    display("Prompt:", question, Warning, HighPriority)
-    displayCategory("Answer:", Warning, HighPriority)
-    let user = stdin.readLine()
-    if user.len == 0: return promptCustom(question, default)
-    else: return user
+  return promptCustom(dontForcePrompt, question, default)
+
+proc promptList*(forcePrompts: ForcePrompt, question: string, args: openarray[string]): string =
```
dom96: This is great but what I had in mind was being able to make a selection using the arrow keys. I guess you might implement that later though.

dom96:
> In any case, if you want I can push those changes to this branch and you can review them together or I can hold off.

Hrm, yeah, might be better for another PR at this point.

src/nimble.nim
```diff
@@ -693,73 +693,126 @@ proc dump(options: Options) =
...
+
+  # Create source directory
+  os.createDir(pkgSrcDir)
+
+  display("Success:", "Source directory created successfully", Success,
+    MediumPriority)
+
+  # Create initial source file
+  cd pkgSrcDir:
+    pkgName.changeFileExt("nim").writeContents "echo \"Hello, World!\"\n"
```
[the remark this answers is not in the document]
antizealot1337: I agree and intended for the default to be library. I just didn't know what else to write. Could write a comment preamble like:
```nim
# pkgName
# Copyright pkgAuthor
# pkgDescription
```
Question about prompting, is there enough difference between *lib*, *bin*, and *lib+bin* packages to warrant the use of a selection, or is *bin* and *lib+bin* similar enough that a simple yes/no prompt is acceptable?

src/nimble.nim
```diff
@@ -693,73 +693,126 @@ proc dump(options: Options) =
...
+    MediumPriority)
+
+  cd pkgTestDir:
+    "tester.nims".writeContents("""switch("path", "$$projectDir/../$#")""" %
+      [pkgSrcDir])
+
+    display("Success:", "Test config file created successfully", Success,
+      MediumPriority)
+
+    "tester.nim".writeContents("doAssert(1 + 1 == 2)\n")
```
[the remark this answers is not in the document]
antizealot1337: I was actually not aware of that. Thanks. I'll change it

src/nimblepkg/tools.nim
antizealot1337: Okay. I'll change to use `writeFile` and delete it

src/nimble.nim
```diff
@@ -693,73 +693,126 @@ proc dump(options: Options) =
...
+  let pkgDesc = promptCustom(options, "Package description?",
+    "A new awesome nimble package")
 
   # Ask for license
-  # TODO: Provide selection of licenses, or select random default license.
-  let pkgLicense = promptCustom("Package license?", "MIT")
+  let pkgLicense = options.promptList("Package License?", [
+    "MIT",
+    "BSD2",
+    "GPL3",
```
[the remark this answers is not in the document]
antizealot1337: Okay. Intended to see what the consensus was from the existing packages but got sidetracked. Will update.

dom96:
Brilliant, thanks! Another nice thing would be to explain what a binary and library package entails in the list selection, if you could fit that in that would be nice.

PR #441, opened 2017-12-22
Prompt for multiple selections interactively when possible

antizealot1337:
This provides a selection interface when it can and preserves the previous behaviour when it cannot.

A few things to note:
* It will only display the selection instructions once since it is a `Warning`.
* `styleDim` on Windows doesn't look very dim. I considered not using it since it will be a different experience on different platforms but left it for now.

Let me know about anything that needs changing (including the package type descriptions). :-)

### DIFF
--- a/src/nimblepkg/cli.nim
+++ b/src/nimblepkg/cli.nim
@@ -213,9 +213,9 @@ proc getKey(): Key =
     else: discard
     result = getKey()
 
-proc promptListInteractive(question: string, args: openarray[string]) : string =
+proc promptListInteractive(question: string, args: openarray[string]): string =
   display("Prompt:", question, Warning, HighPriority)
-  display("Select", "With up/down arrow keys, 'Enter' when done", Warning,
+  display("Select", "With up/down arrow keys, 'Enter' when done", Message,
     HighPriority)
   displayCategory("Choices:", Warning, HighPriority)
   var
--- a/src/nimblepkg/cli.nim
+++ b/src/nimblepkg/cli.nim
@@ -12,7 +12,10 @@
 #   - Bright for HighPriority.
 #   - Normal for MediumPriority.
 
-import logging, terminal, sets, strutils
+import logging, terminal, sets, strutils, os
+
+when defined(windows):
+  import winlead
 
 type
   CLI* = ref object
@@ -33,6 +36,9 @@ type
 
   ForcePrompt* = enum
     dontForcePrompt, forcePromptYes, forcePromptNo
+
+  Key = enum
+    kNone, kArrowUp, kArrowDown, kEnter
 
 const
   longestCategory = len("Downloading")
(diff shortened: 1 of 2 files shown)


src/nimblepkg/cli.nim
```diff
@@ -169,20 +175,124 @@ proc promptCustom*(forcePrompts: ForcePrompt, question, default: string): string
 proc promptCustom*(question, default: string): string =
   return promptCustom(dontForcePrompt, question, default)
 
+proc getKey(): Key =
+  ## Needed in order to allow arrow key control for Windows. It also makes it
+  ## easier to determine which keys are pressed on POSIX systems. This is based
+  ## of terminal.getch for the standard library.
```
dom96: Why can't this be used directly?

src/nimblepkg/cli.nim
```diff
@@ -169,20 +175,124 @@ proc promptCustom*(forcePrompts: ForcePrompt, question, default: string): string
...
+        case getch():
+        of 'A':
+          return kArrowUp
+        of 'B':
+          return kArrowDown
+        else: discard
+    else: discard
+    result = getKey()
+
+proc promptListInteractive(question: string, args: openarray[string]) : string =
```
dom96: Style nitpick: no space before ``:``.

src/nimblepkg/cli.nim
```diff
@@ -169,20 +175,124 @@ proc promptCustom*(forcePrompts: ForcePrompt, question, default: string): string
...
+  display("Prompt:", question, Warning, HighPriority)
+  display("Select", "With up/down arrow keys, 'Enter' when done", Warning,
```
dom96: Yeah, I think it would be better if this isn't a Warning.

dom96:
Thank you for working on this :)

src/nimblepkg/cli.nim
antizealot1337: I initially tried that but under Windows `getch` returns only `0` with the arrow keys (works find for enter and space). The standard library returns `keyEvent.uChar` which doesn't appear to hold an actual value when the arrow keys are pressed. Using `getch` alone would prevent Windows users from selecting with the arrow keys. Could use the space to cycle through items on Windows. I thought about changing/adding to `terminal`, but I'm unsure what would be the best way forward.

src/nimblepkg/cli.nim
antizealot1337: My preference as well. I guess I just hit the space bar an extra time

src/nimblepkg/cli.nim
antizealot1337: I thought it would work better as a `Message` but left it as Warning for consistency. I'll change it.

src/nimblepkg/cli.nim
dom96: The best thing would be to patch the stdlib I think for this case.

dom96: But that can be separately if you wish.

antizealot1337: I think you're right. I submitted nim-lang/nim#6966 to attempt to fix it. If it gets accepted `getKey` may still be useful but can be removed along with the Key enum.

dom96:
So https://github.com/nim-lang/Nim/pull/6966 is keeping this in limbo I think. Any thoughts on what we should do about it?

antizealot1337:
It is.  I can think of a few ways to move this forward:

1. Move forward without nim-lang/Nim#6966 by a either:
  a. Keeping it as is (Not preferred. I think nim-lang/Nim#6966 will eventually be resolved and there will be artifacts in the `cli` module that will need to be removed)
  b. Move all the `getKey` and the `Key` enum to a separate module so when nim-lang/Nim#6966 is resolved it can be yanked out with little change. (Doesn't really fix the problem but may make it easier to remove later)
2. Wait for nim-lang/Nim#6966 to be resolved and then fix this to use it instead (or implement it from scratch using the resolution of nim-lang/Nim#6966).

I'm not partial to any option. I can say that keeping it around waiting for the resolution of the nim issue may make it harder to fix later. I checked out a newer compiler than `stable` and the changes to terminal have created some issues that I haven't looked into yet (don't know if the changes exposed a bug in my code or if my code exposed a bug in `terminal` or something else). Also, I only tested on macOS so no clue how it acts Linux or Windows. I'll dig into the problem and try to get it fixed. After I get that I can rebase and push code and fix anything else to move this forward.

antizealot1337:
Hey @dom96 this has been open for a while and I've had an idea on how this can use `terminal.getch` and still be cross platform without the dirty hack that's in there now. What if the tab and/or space were used to cycle through options? Without a comprehensive cross platform key API I don't know what else to do on this front. Thoughts?

dom96:
Sorry this took so long!

antizealot1337:
No worries :)

treeform:
This does not work on windows for me.

antizealot1337:
Hi, sorry for the late reply.

After a little experimentation I found that nimble will fallback to
the old input using Bash both from Cygwin and the one installed along
with Git but not with cmd.exe on Windows 7. I have not tried
PowerShell as of yet. This fallback behavior means the terminal
packages isatty is returning false. I don't know why this is. I'll
look into this more and see if I can find what the problem is.

With that said, are you experiencing the "fallback" behavior or something else?

PR #443, opened 2017-12-22
Fix deprecated warning for getBVal

Yardanico:


Yardanico:
@dom96 well it still doesn't pass :)
I guess we'll really have to wait until new Nim release arrives

dom96:
It's better to keep this for now anyway to support older versions of Nim.

I think the reason travis fails is because it uses Nim 0.17.0 to test Nimble.

Issue #280, opened 2016-11-25
`before install` not called when installing from nim-lang/packages


vktec:
When running `nimble install` within my package directory, `before install` blocks are executed as normal. When installing from nim-lang/packages with `nimble install mypackage`, they are not.

I'm not sure how to gather more information about this, as I can't choose an alternate source to install from, as far as I can tell.

vktec:
Installing directly from Git using `nimble install https://foo/bar.git` works as expected,

vktec:
Ah, that's wrong. I was running the command from within my project directory, which made it succeed.

vktec:
I think I've tracked down the issue.

When Nimble starts, it first parses command line arguments, then runs the `doAction` proc on the resulting `Options` object. One of the first things `doAction` does is to run `execHook`, which then looks for a `.nimble` file in the current directory. If it can't find one, it exits without doing anything. This is why it works as expected when run in my project directory, but not when running elsewhere.

The problem is that fixing this appears like it would require a rather major refactor of Nimble, so I think it's better for me to work around this issue by creating a separate task which I can run every so often to update my wrapper, rather than doing it directly before install.

dom96:
This is still an issue though. Doesn't matter how difficult it is to fix.

genotrance:
Unfortunately this is affecting [nimssl](https://github.com/genotrance/nimssl) and [nimbass](https://github.com/genotrance/nimbass) as well since they actually generate the wrappers using [nimgen](https://github.com/genotrance/nimgen) before installing to the ~/.nimble directory.

I've updated my README as well like khogeland's commit above but a fix would be much cooler.

dom96:
@genotrance cool library, I'll make this a high priority.

dom96:
This was fixed it seems :)

dom96:
Reopening as this wasn't fixed properly.

dom96:
So just to clarify, this was fixed, but the hook is run too quickly (before dependencies are installed), so you get issues like this when installing packages that depend on 'nimgen' (as an example):

<img width="965" alt="screen shot 2018-03-18 at 01 14 24" src="https://user-images.githubusercontent.com/246651/37561608-ae731b00-2a4a-11e8-901d-5014a5da6ff8.png">

genotrance:
Yes exactly - the latest commits fixes this exact problem.

PR #448, opened 2018-01-08
Fix for #280

genotrance:
I've fixed this by calling ```execHook()``` within install() again. It is the only special case that the hooks need to be called again since the package artifacts weren't available when the usual checks were performed for hooks.

The existing calls to ```execHook()``` are also in the right place since almost every other nimble sub-command is expected to run on an existing nimble project directory.

Note that this fix will only run before and after ```install``` hooks and not any before and after ```build``` hooks (install calls build internally) since, as should be expected, the sub-command was ```install``` and not ```build``` so running the build before/after hooks isn't really warranted. That being said, it should be clarified in the documentation so that expectations are clear.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -468,7 +468,17 @@ proc install(packages: seq[PkgTuple],
       let (downloadDir, downloadVersion) =
           downloadPkg(url, pv.ver, meth, options)
       try:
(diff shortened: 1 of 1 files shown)


dom96:
Nice! Seems like an attractive way to fix this issue, thanks :)

Issue #398, opened 2017-08-24
publishing fails with bad error report


krux02:
This error message is not very informative, please write something that help to detect what I actually did wrong.

```
$nimble publish
       Info Hint: used config file '/home/arne/proj/nim/Nim/config/nim.cfg' [Conf]
      Info: Using GitHub API Token in file: /home/arne/.nimble/github_api_token
   Success: Verified as krux02
    Cloning packages into: /tmp/nimble-packages-fork
   Updating the fork
    Prompt: Whitespace separated list of tags?
    Answer: gui rendering opengl
    Pushing to remote of fork.
       Info Creating PR
Error: unhandled exception: Connection was closed before full request has been made [ProtocolError]
```

dom96:
This seems to be a recent regression. I'm guessing GitHub changed something :\

edubart:
I'm also getting this error

bogen:
I just now tried to publish 2 packages, same error.

Gooseus:
I'm also having this issue trying to publish a package - nimble v0.8.8

genotrance:
Since nothing was being pushed (due to nil pw used in push step), the PR failed (there was nothing to pull on remote. I've pushed a fix along with proxy support for publish.

PR #451, opened 2018-01-14
Fix for #398

genotrance:


dom96:
git over https won't work with SSH keys AFAIK (which is how most people's git is configured including mine).

I don't understand why this ``pw`` thing was there in the first place and how it worked in the past.

genotrance:
Given ```nimble publish``` requires an access token for all http API requests anyway, I've updated the code to simply reuse the same access token for pushing over http as well. This avoids the need to rely on any existing SSH key configuration and more proxy friendly in general.

Apart from this, I'm no longer cloning the repo but instead doing:
```
git init
git pull https://github.com/gituser/packages
```
This is per recommendation from [this](https://github.com/blog/1270-easier-builds-and-deployments-using-git-over-https-and-oauth) Github blog post to avoid writing the access token into the forked packages repo.

dom96:
Awesome, didn't know it was possible to use tokens this way.

PR #454, opened 2018-01-30
Fixed nimble path

yglukhov:
2 issues:
1. `nimble path foo` would return a path for `foobar` package (same prefix).
2. `nimble path foo` returns wrong path if multiple versions are available.


dom96:
Thanks. I'll accept this but I think in the long run this proc need to be rewritten to use Nimble's package handling procs.

yglukhov:
Thank you!

PR #463, opened 2018-03-01
Fortify Nimble build

FedericoCeratto:


dom96:
This should be tested on Windows.

FedericoCeratto:
The set of flags will protect the executable from statically linked C libraries (that might be used in future).

genotrance:
I don't think there's any plans to statically link libs at this point. Nimble will always live with Nim which carries several dynamic libs. I propose closing.

PR #464, opened 2018-03-11
Updated fix for #280 with dependencies

genotrance:
Original fix ran "before install" before dependencies were installed.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -348,6 +348,7 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
     return result
 
   # Run pre-install hook in download directory now that package is downloaded
+  # and all dependencies have been installed
   cd dir:
     if not execHook(options, true):
       raise newException(NimbleError, "Pre-hook prevented further execution.")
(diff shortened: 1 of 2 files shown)


genotrance:
I don't see why #280 got linked to these unrelated comments. Regardless, this updated fix has a new issue - if you run `nimble install` on a package in a local directory, it runs `before install` twice since the execHook() runs in doAction() as well as inside install().

I'm thinking we need to keep track of whether before and after hooks have run already or not. I think this can be done by changing PackageInfo.preHook and postHook into Table[string, bool] but it might be overkill. Suggestions appreciated.

genotrance:
I've fixed this by moving the pre and post execHook() calls within installFromDir(). This ensures that it works correctly whether installing locally or after download.

Additionally, I prevent the generic execHook() calls in doAction() from running for actionInstall, similar to actionCustom since they are handled elsewhere.

dom96:
Relevant issue: https://forum.nim-lang.org/t/3949 (which we should also fix)

dom96:
Closing in favour of #482.

Issue #393, opened 2017-08-16
Don't turn .nimble-link pkg info into full pkg infos


dom96:
See TODO in nimble.nim line ~245 (search for '#393')

Perform the logic in `getInstalledPkgsMin`.

PR #465, opened 2018-03-15
Fix wrong version detection in listPaths.

BopoHDark:
In case we have packages with same name and aone of them has nimble-link(as result of nimble develop i.e.), version of package with nimble-link will be the same as version of package which that link targets.This leads to wrong nimble path results when you have developed and installed versions of the package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,16 @@ language: c
 
 install:
   - |
-      wget https://nim-lang.org/download/nim-0.17.0.tar.xz
-      tar -xf nim-0.17.0.tar.xz
-      cd nim-0.17.0
+      wget https://nim-lang.org/download/nim-0.17.2.tar.xz
+      tar -xf nim-0.17.2.tar.xz
+      cd nim-0.17.2
       sh build.sh
       cd ..
 
 before_script:
   - set -e
   - set -x
-  - export PATH=`pwd`/nim-0.17.0/bin:$PATH
+  - export PATH=`pwd`/nim-0.17.2/bin:$PATH
 
 script:
   - cd tests
--- a/changelog.markdown
+++ b/changelog.markdown
@@ -3,6 +3,38 @@
 
 # Nimble changelog
 
+## 0.8.10 - 23/02/2018
+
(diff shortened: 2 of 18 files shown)


dom96:
Yay! Beautiful! Thank you for using the ``getInstalledPkgsMin`` function, should have been done in the first place :)

dom96:
If you could resolve the conflicts I will merge it immediately. Otherwise will wait until I'm working on the next release.

BopoHDark:
Done. Sorry couldn't do this earlier.

BopoHDark:
One of tests of *[Suite] develop feature*, test "can develop with srcDir" is broken. I don't know how to fix it, it calls develop in directory "develop/srcdirtest", and after that checks path "develop/srcdirtest/src".

dom96:
Well... it was passing before. Do you think the test is incorrect?

[Link to test failure](https://travis-ci.org/nim-lang/nimble/builds/355320410#L909)

BopoHDark:
Problem was in srcDir field witch is "" for package which isLinked, so getRealDir does not return proper path.
I've tryed to workaround this, but after my commit tests are failed, when my local tests is ok. :(

dom96:
No worries, I'll look into it later.

Vindaar:
I rebased this PR on the current master locally to check whether it solves a problem I had: 
Running `nimble path` on an installed package returned the `srcDir` appended to the path, even though it shouldn't. The reason was that `isLinked` and `isInstalled` are not set in the current master ins `listPaths`. In line https://github.com/nim-lang/nimble/blob/master/src/nimble.nim#L640 the call to `getRealDir` then returns the `srcDir` appended, since `isInstalled` is falsely false.

Since this PR uses `getInstalledPkgsMin` in `listPaths` instead of walking the directory, `isInstalled` and `isLinked` (thanks to the introduced `getLinkedSrcDir`) are properly set and the `srcDir` doesn't show up.

I ran `nimble test` locally and do not see any errors. 

And if I'm not mistaken the issue #393 / lines https://github.com/nim-lang/nimble/blob/master/src/nimble.nim#L193-L196 can now be fixed with this too, since `srcDir` is now properly set.

PR #466, opened 2018-03-20
binDir

shinriyo:
binDir prompt added


dom96:
I don't think ``binDir`` is important enough to have a prompt. Sorry.

shinriyo:
sign sign

PR #467, opened 2018-03-20
for a sense of unity.

shinriyo:
another is .nimble. so I modified.


dom96:
Upvoting your own PR won't make it more likely to be merged ;)

The current error message is fine. What would be a welcome improvement is appending the filename of the Nimble file: ``Nimble file already exists: /home/user/projects/foo/foo.nimble``

shinriyo:
@dom96 
I did your feedback.

dom96:
Still not 100%, but I made the necessary changes.

shinriyo:
@dom96 yes it is better.

PR #494, opened 2018-05-21
Add note about SSH key to the section on nimble publish

Aearnus:
This was an issue I was trying to figure out for a solid few minutes, and it wasn't mentioned anywhere else as far as I could find.


dom96:
Nice! Thanks for creating this PR :)

PR #512, opened 2018-07-13
Add missing call to cleanup()

LemonBoy:


dom96:
If you're up for it, could we get a test for this too?

PR #528, opened 2018-08-28
check package name conflict locally

bung87:
background story:
when the first time I publish a package,It created a PR, test through CI service after it build several minutes told me my package name conflict with others :(


dom96:
That's a really large function for just checking for duplicates...

bung87:
but Think it may response me ASAP that package name existed ,before created a PR ,and building through CI ,minutes late.

PR #532, opened 2018-09-05
Fixes #503 by creating directory if current and name differ

### LINKED ISSUE #503: Make nimble init generate the Nimble project dir too
Right now, `nimble init foo` generates `src/foo.nim`, etc. directly in the current directory.

I would expect that to generate those but **first create a `foo` directory** and then generate those inside that.

https://gitter.im/nim-lang/Nim?at=5b22b38dc277fb705dce91fa

antizealot1337:
An attempt to fix #503 without adding new flags or commands. If a package name is provided to `init` This will change attempts to _do the right thing_ depending on if the package name and the current directory are the same. If they are not a sub directory is created to contain the package files. If they are the current directory is populated with the package files.

If no package name is provided the behaviour is the same as before.

This implementation prevents one current behaviour. Currently nimble will create a package with a name different from the directory name if a name is provided and it is different.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -698,25 +698,31 @@ proc dump(options: Options) =
   echo "backend: ", p.backend.escape
 
 proc init(options: Options) =
-  var nimbleFile: string = ""
-
   if options.forcePrompts != forcePromptYes:
     display("Info:",
       "In order to initialise a new Nimble package, I will need to ask you\n" &
       "some questions. Default values are shown in square brackets, press\n" &
       "enter to use them.", priority = HighPriority)
 
-  # Ask for package name.
+  # Determine the package name.
   let pkgName =
     if options.action.projName != "":
       options.action.projName
     else:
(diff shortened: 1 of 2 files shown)


src/nimble.nim
```diff
@@ -698,25 +698,31 @@ proc dump(options: Options) =
...
     else:
       os.getCurrentDir().splitPath.tail.toValidPackageName()
 
-  nimbleFile = pkgName.changeFileExt("nimble")
-  validatePackageName(nimbleFile.changeFileExt(""))
+  # Validate the package name.
+  validatePackageName(pkgName)
+
+  # Determine the package root.
+  let pkgRoot = if pkgName == os.getCurrentDir().splitPath.tail:
```
dom96: Small nitpick for future reference: if should be on its own line.

antizealot1337: Gotcha. I was going to fix it with another pull request to add more licenses but I see that's taken care of already. :)

PR #537, opened 2018-09-10
make Nimble more robust for AppVeyor

Araq:


Araq:
No, it can raise in general and it's not a critical error for Nimble at all.

PR #541, opened 2018-09-14
Add a few useful os calls to nimble

genotrance:
Nimble equivalent of the `setupVM()` enhancements in https://github.com/nim-lang/Nim/pull/7442.


kaushalmodi:
Ref: https://gitter.im/nim-lang/Nim?at=5b9a8d51b4990c30eeadb6a5

dom96:
It would indeed be nice if we didn't need to duplicate this from Nim.

kaushalmodi:
@dom96 Can this PR be merged?

genotrance:
Fixed the conflict - should be good to go for now. I'll look into leveraging nimscript directly in the future.

PR #554, opened 2018-10-10
Additional information on Nim compiler integration

technicallyagd:
It took me some trial and error to figure out how `nimblePath` works, so I thought this bit of info may help others in the future. I am actually not sure whether this clarification belongs here or in Nim's compiler documentation, please let me know if it's the latter.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -748,14 +748,14 @@ to ensure that the dependencies you specified are correct.
 
 ### Compile with `nim` after changing the nimble directory
 
-`nim` has been preconfigured to look at the default nimble directory while compiling,
+The Nim compiler been preconfigured to look at the default nimble directory while compiling,
 so no extra step is required to use nimble managed packages in your code. 
-However, if you are using custom `nimbleDir`, you need to specify 
-`--NimblePath:PATH` option (`nimblepath` is case insensitive). For example,
+However, if you are using a custom `nimbleDir`, you need to specify 
+`--nimblePath:PATH` option. For example,
 if your `nimble` directory is located at `/some/custom/path/nimble`, this should work:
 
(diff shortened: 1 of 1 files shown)


readme.markdown
```diff
@@ -746,6 +746,23 @@ This means that you can safely compile using the compiler when developing your
 software, but you should use Nimble to build the package before publishing it
 to ensure that the dependencies you specified are correct.
 
+To compile code that uses nimble managed packages with `nim`,you need
```
dom96: This is good but IMO it should be under its own section called "Changing the default Nimble dir" or something.

It should emphasise that you shouldn't need this normally because ``~/.nimble`` is already added to Nim's ``nimblePath`` by default.

readme.markdown
```diff
@@ -746,6 +746,23 @@ This means that you can safely compile using the compiler when developing your
...
+to specify `--NimblePath:PATH` option. For example, if your `nimble` directory
```
dom96: This isn't capitalised AFAIK

readme.markdown
technicallyagd: Yeah, I ran into this problem because I had custom nimble path. I will emphasize this in next revision.

readme.markdown
technicallyagd: On the [documentation](https://nim-lang.org/docs/nimc.html#compile-pragma), it's PascalCased, but I think it's case insensitive.

technicallyagd:
I have made the modifications you requested, @dom96 . Just in case Github didn't notify you earlier.

dom96:
Thanks!

technicallyagd:
My pleasure :D

Issue #555, opened 2018-10-11
Add a `nimble clean` command to clean up build files.


davidgarland:
Having a built-in `nimble clean` command that cleans up any test or binary files that have been built via for instance `nimble build` would be nice. Not quite as pressing of an issue now that nimcache in the local directory isn't a thing, but it would still be a huge improvement over having to write

```Nimble
task clean, "Cleans up files":
  exec "rm <project_binary_name_here>"
```

for every project.

jabbalaci:
I just noticed that "nimble test" creates a binary file for every test file in the `tests/` folder. It would be nice if `nimble test` could do a cleanup, for instance via an option: `nimble test --clean`.

epasveer:
The latest "pull" works well for me.

PR #566, opened 2018-10-19
#555: `nimble test` removes the created binaries

jabbalaci:


src/nimble.nim
```diff
@@ -1002,7 +1002,17 @@ proc test(options: Options) =
       optsCopy.action.compileOptions = @[]
       optsCopy.action.compileOptions.add("-r")
       optsCopy.action.compileOptions.add("--path:.")
+      let
+        binFileName = joinPath(dir, name) & (if defined(windows): ".exe" else: "")
```
dom96: This can be simplified to:

```
binFileName = file.path.changeFileExt(ExeExt)
```

dom96: btw, use ``/`` instead of ``joinPath`` for future reference.

jabbalaci:
I pushed another commit and changed that line. However, I might have messed up the "resolve conversation".

dom96:
Not sure what you mean, but looks good.

PR #588, opened 2018-11-28
Don't swallow control-C

### LINKED ISSUE #587: promptListInteractive swallows Control-C
If you run the following nim program:

```nim
import nimblepkg/cli
discard promptList(dontForcePrompt, "Choose something?", ["a", "b", "c"])
```

And press Control-C, it will *not* break out of the prompt, but I expect that it *should* stop the program.

I'm on macOS.

```
nimble v0.9.0 compiled at 2018-10-19 22:47:19
git hash: 22aa3c1d1392d7878d039250ecf0ffeadf3ba0cc
```

iffy:
Fixes #587 

But doing `quit(1)` seems like the wrong thing to do.  Is there a better way to throw the error that Control-C normally causes?

### DIFF
--- a/src/nimblepkg/cli.nim
+++ b/src/nimblepkg/cli.nim
@@ -13,7 +13,7 @@
 #   - Normal for MediumPriority.
 
 import logging, terminal, sets, strutils, os
-import ./version
+import ./common
 
 when defined(windows):
   import winlean
--- a/src/nimblepkg/cli.nim
+++ b/src/nimblepkg/cli.nim
@@ -13,6 +13,7 @@
 #   - Normal for MediumPriority.
 
 import logging, terminal, sets, strutils, os
+import ./version
 
 when defined(windows):
   import winlean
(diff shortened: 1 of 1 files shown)


src/nimblepkg/cli.nim
```diff
@@ -218,6 +218,9 @@ proc promptListInteractive(question: string, args: openarray[string]): string =
       of '\r':
         selected = true
         break
+      of '\3':
+        showCursor(stdout)
+        quit(1)
```
dom96: You can use ``QuitFailure`` here instead. But what you should really do is raise some exception, ``NimbleError`` should be fine for this.

iffy: k, this version throws a NimbleError instead.

src/nimblepkg/cli.nim
```diff
@@ -13,6 +13,7 @@
 #   - Normal for MediumPriority.
 
 import logging, terminal, sets, strutils, os
+import ./version
```
dom96: Why is this required? Should it be "common"?

iffy: I just grepped for the definition of `NimbleError` and found it in version :)  I'll change it to `common`

PR #623, opened 2019-03-29
Update packageparser.nim

lihf8515:
The next version of the `parsecfg` module will support empty rows and comments, resulting in a situation where values are empty.

### DIFF
--- a/src/nimblepkg/packageparser.nim
+++ b/src/nimblepkg/packageparser.nim
@@ -259,6 +259,10 @@ proc readPackageInfoFromNimble(path: string; result: var PackageInfo) =
             case result.backend.normalize
             of "javascript": result.backend = "js"
             else: discard
+          of "": # The next version of the `parsecfg` module will support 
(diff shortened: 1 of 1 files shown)


dom96:
Wat. Link?

lihf8515:
"parsecfg.nim" I canundefinedt submit now because the "packageparser.nim" of "nimble" does not deal with the situation where the read key is empty. Therefore, you need to deal with the situation where the key is empty first. Then I can submit "parsecfg.nim" correctly.

lihf8515:
"parsecfg.nim" and "packageparser.nim" are interdependent, and to successfully submit a "parsecfg.nim" that supports comments and empty lines, you must first let "packageparser.nim" handle it. For more information, see # 623

lihf8515:
https://github.com/nim-lang/Nim/pull/10928

dom96:
Please don't close and reopen multiple PRs like this.

Issue #640, opened 2019-04-27
nimble dump wants an answer


FedericoCeratto:
Nimble dump gets stuck asking for an answer when run against https://github.com/FedericoCeratto/nim-syslog/ at tag 0.0.3

```
$ nimble dump 
    Answer: 
```

nimble dump -n  prints nothing
nimble dump -y  prints the expected dump

jrfondren:
For reference this is the .nimble file that caused the problem:
```
[Package]
name          = "syslog"
version       = "0.0.3"
author        = "Federico Ceratto"
description   = """Minimal syslog support."""
license       = "LGPLv3"

installDirs = """


"""

InstallFiles = """

LICENSE
README.md
syslog.nim

"""

[Deps]
Requires: """

nake >= 1.4
nim >= 0.10.2

"""
```

genotrance:
Is this fixed in #648?

PR #641, opened 2019-04-28
Option to continue tests on failure

jrfondren:
- add -c/--continue flags to "nimble test", to prevent it from exiting on the first test failure
- at the end of a test run, if there were failures, show an error with the success rate

A near equivalent is the following custom task:

```nim
task test, "Run test suite":
  exec """for x in tests/t*.nim; do nim c --path:"../src" -r $x; done"""
```

Abbreviated example output with these patches:

```
# ../nimble/nimble test -c
  Executing task test in /Users/jfondren/nim/modsec/modsec.nimble
  Verifying dependencies for modsec@0.1.0
      Info: Dependency on npeg@>= 0.10.0 already satisfied
  Verifying dependencies for npeg@0.10.0
  Compiling /Users/jfondren/nim/modsec/tests/test_actions_oddities.nim (from package modsec) using c backend
[FAILED] can parse actions that end in a comma
[FAILED] can parse actions that end in a stray single quote
Error: execution of an external program failed: '/Users/jfondren/nim/modsec/tests/test_actions_oddities '
  Verifying dependencies for modsec@0.1.0
      Info: Dependency on npeg@>= 0.10.0 already satisfied
  Verifying dependencies for npeg@0.10.0
  Compiling /Users/jfondren/nim/modsec/tests/test_normal.nim (from package modsec) using c backend
[OK] can parse rule with escaped quote in string
...
[OK] can parse rule with unicode double-quotes within double-quoted strings
   Success: Execution finished
     Error: Only 2/3 tests passed
```

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -677,9 +677,15 @@ proc getPackageByPattern(pattern: string, options: Options): PackageInfo =
       )
     result = getPkgInfoFromFile(skeletonInfo.myPath, options)
 
+proc projName(options: Options): string =
+  if options.action.projName != "":
+    options.action.projName
+  else:
+    os.getCurrentDir().splitPath.tail.toValidPackageName()
+
 proc dump(options: Options) =
   cli.setSuppressMessages(true)
-  let p = getPackageByPattern(options.action.projName, options)
+  let p = getPackageByPattern(projName(options), options)
   echo "name: ", p.name.escape
   echo "version: ", p.version.escape
   echo "author: ", p.author.escape
@@ -699,11 +705,7 @@ proc dump(options: Options) =
 
 proc init(options: Options) =
   # Determine the package name.
-  let pkgName =
-    if options.action.projName != "":
-      options.action.projName
-    else:
-      os.getCurrentDir().splitPath.tail.toValidPackageName()
+  let pkgName = projName(options)
 
   # Validate the package name.
   validatePackageName(pkgName)
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -992,7 +992,9 @@ proc develop(options: Options) =
 proc test(options: Options) =
   ## Executes all tests starting with 't' in the ``tests`` directory.
   ## Subdirectories are not walked.
-  var files = toSeq(walkDir(getCurrentDir() / "tests"))
+  var
+    files = toSeq(walkDir(getCurrentDir() / "tests"))
+    tests, failures: int
 
   if files.len < 1:
     display("Warning:", "No tests found!", Warning, HighPriority)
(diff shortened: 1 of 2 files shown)


jrfondren:
NB. the 2/3 refers to the test.nim files that were executed, not the individual tests.

I think some closer integration with the unittest module could be useful, but this PR doesn't go that far.

bluenote10:
That's also the behavior I would prefer as default. However I think it should be combined with hiding the output of successful tests, otherwise it can be difficult to identify the stdout from the failed tests from all the non-informative successful tests (since the Nim compilation can be somewhat verbose, the total output running multiple tests can be quite long). An example of that can be seen [here](https://travis-ci.org/bluenote10/oop_utils/builds/521431049#L1966) using [this implementation](https://github.com/bluenote10/oop_utils/blob/master/oop_utils.nimble#L68).

jrfondren:
@bluenote10 you can add --hints:off to your tests/nim.cfg to hide a lot of that.

bluenote10:
I know, but I want them in general. It's just that the output of successful tests is insignificant compared to the output of failed tests. Hiding details from successful tests is the default behavior in all major test runners.

dom96:
Yeah. Nimble should hide compilation output by default, you can easily make it call Nim with `--hints:off`.

src/nimble.nim
```diff
@@ -677,9 +677,15 @@ proc getPackageByPattern(pattern: string, options: Options): PackageInfo =
...
+proc projName(options: Options): string =
+  if options.action.projName != "":
+    options.action.projName
+  else:
+    os.getCurrentDir().splitPath.tail.toValidPackageName()
+
 proc dump(options: Options) =
   cli.setSuppressMessages(true)
-  let p = getPackageByPattern(options.action.projName, options)
+  let p = getPackageByPattern(projName(options), options)
```
dom96: Why use that here? `nimble dump` works fine, no?

jrfondren: @dom96 , that's a fix for #640 , where dump was hanging on an invalid .nimble file. I'm new to git, and only after this started making PRs from branches dedicated to different issues.

dom96: Yeah, please separate these next time. I sadly don't understand your fix so I'm not certain enough to accept it, is it something about the package in #640 that causes that bug? I've ran `nimble dump` on master and it worked fine for Nimble itself.

jrfondren: nim-syslog's already fixed its .nimble file. I added a comment to #640 with the .nimble that triggered the problem. With that .nimble file, `options.action.projName` was the empty string.

dom96: What is it about that Nimble file that causes this?

jrfondren: I'm not actually sure, but I suspect the name in the file itself.

It'll be a few days before I can get back around to this, but I'll follow up.

PR #642, opened 2019-04-30
Fix quoted switch, multi-line description, more caching

genotrance:


Araq:
The (un)escape could be cleaner but I consider this a hotfix and so it needs to be merged asap.

PR #644, opened 2019-05-02
Template support for Nimble

nais314:
Extended nimble with the **sample** command (as template is a reserved word).
Frameworks may add Application Templates to the nimble package, like SimpleApp, AdvancedApp, etc,
or Demo applications may be shipped with the package.
    ~/myproject $ nimble sample [pkgname] [samplename]
    ~/myproject $ nimble sample [pkgname] [pkgversion] [samplename]
    ~/myproject $ nimble sample [pkgname] 
Added **samples** directory to the folder structure.

Hope you like it!
Best Wishes:
István


nais314:
Extended nimble with the **sample** command (as template is a reserved word).
Frameworks may add Application Templates to the nimble package, like SimpleApp, AdvancedApp, etc,
or Demo applications may be shipped with the package.
~/myproject $ nimble sample [pkgname] [samplename]
~/myproject $ nimble sample [pkgname] [pkgversion] [samplename]
~/myproject $ nimble sample [pkgname]
Added samples directory to the folder structure.

Hope you like it!
Best Wishes:
István

nais314:
fixed: ~/myproject $ nimble sample [pkgname] [pkgversion]

readme.markdown
```diff
@@ -310,6 +311,17 @@ Outputs information about the package in the current working directory in
...
 
+### nimble sample
+
+The ``sample`` command copies from the packages samples directory to the
+working directory. (It does not creates new directory, and it overwrites 
+existing files)
+
+    myproject$ nimble sample [pkgname]
+    myproject$ nimble sample [pkgname] [samplename]
+    myproject$ nimble sample [pkgname] [pkgversion] [samplename]
```
dom96: Package names and versions have a native syntax in Nimble: `pkg@0.1.0`. You should support this in this command instead of separating by a space.

readme.markdown
```diff
@@ -480,6 +492,11 @@ For a package named "foobar", the recommended project structure is the following
     ├── nim.cfg
     ├── tfoo1.nim   # First test
     └── tfoo2.nim   # Second test
+└── samples
+    ├── demo1       # Demo application
+    ├── simpleapp   # Application template
+    └── foo
+        └── subfoo  # can be nested
```
dom96: The lack of file extension here implies that these are all binaries

readme.markdown
dom96: I'm also not sure how intuitive this command is, how does the user know what samples are available? Does the user look it up in the package's source code? If so, at that point it might just be easier to copy the file manually.

Perhaps a better approach would be (in general I would prefer the "example" nomenclature):

* ``nimble examples pkgname`` - displays a list of examples for `pkgname`
* ``nimble examples pkgname foobar`` - copies `foobar.nim` from `pkgname`'s examples to `./foobar.nim` (errors if `./foobar.nim` exists)
* ``nimble examples pkgname@0.1 foobar`` - copies `foobar.nim` from `pkgname`'s (at version 0.1) examples to `./foobar.nim` (errors if `./foobar.nim` exists)

What do you think?

src/nimble.nim
```diff
@@ -1043,6 +1043,56 @@ proc check(options: Options) =
...
+
+  for (pkg,meta) in installedPkgs: # pkg exists?
+    if pkg.name == options.action.package and
+      (withinRange(pkg, version) or pkg.version == "#head"): #"#head" == develop 
+        var (dir, name, ext) = splitFile(pkg.myPath)
+        let examplesDir = dir & DirSep & "samples"
+        if not dirExists(examplesDir):
+          display("Problem:", pkg.name & " has no samples.", 
+              priority = HighPriority)
+          raise NimbleQuit(msg: pkg.name & " has no samples.")
```
dom96: Please use `raiseNimbleError` instead of using `display` + raising NimbleQuit.

src/nimble.nim
```diff
@@ -1043,6 +1043,56 @@ proc check(options: Options) =
...
+        #select sample:
+        var 
+          selectedSample: string #this path will be copied
```
dom96: Nit: space after `#`

src/nimble.nim
```diff
@@ -1043,6 +1043,56 @@ proc check(options: Options) =
...
+  ## ~ overwrites existing files without question
+
+  let installedPkgs = getInstalledPkgsMin(options.getPkgsDir(), options)
+  var version = toVersionRange(options.action.ver)
+
+  for (pkg,meta) in installedPkgs: # pkg exists?
+    if pkg.name == options.action.package and
+      (withinRange(pkg, version) or pkg.version == "#head"): #"#head" == develop 
+        var (dir, name, ext) = splitFile(pkg.myPath)
+        let examplesDir = dir & DirSep & "samples"
```
dom96: This directory won't be copied in there. You'll need to implement this installation logic for examples as well (`examples/` should probably go to `~/.nimble/examples/pkg_name-ver`).

src/nimblepkg/options.nim
```diff
@@ -268,8 +278,18 @@ proc parseArgument*(key: string, result: var Options) =
...
+    if result.action.package == "":
+      result.action.package = key
+    elif result.action.sample == "" and 
+      (not ('.' in key)) and (key != "#head"):
+        result.action.sample = key
+    elif (not ('.' in key)) and (key != "#head"):
+      result.action.ver = Version(result.action.sample)
+      result.action.sample = key
+    else:
+      result.action.ver =  Version(key)
```
dom96: yeah, please don't do this. Reuse the logic for parsing versions that I have for `install`.

nais314:
> This is an interesting idea. Is there another package manager that implements something similar?

I don't think so... Frameworks have implemented it individually (eg.: yiic webapp [yourdir]), and for every prog lang in different ways (see Gtk for python or C). 
(But more common is, that you have to copy paste from ReadMe, or download them separately from other repo ). 

A "HelloWorld" app can be found for frameworks, wich is a good starting point (boilerplate).

It would look nice in nim framework readmes, after **Install**: nimble pkgname : 
nimble examples pkgname demo
- and not other multiline instructions for manual copy/paste create etc. ;) - convience.

Well, i could post issues, or try to chat about it, but i think this is almost as good (?)

src/nimble.nim
nais314: (hopefully) we will get back to this, as it is not clear for me, but let me first fix the other issues and add some explanation. thank you!

readme.markdown
nais314: implemented, pushing in a minute

readme.markdown
nais314: these are subfolders.
(eg: multithreading/channels; multithreading/threadpool )

the idea was:
for a framework sometimes directory structures needed with files in them.
polluting project dir with include files is not pretty, include files can go into subfolder,
(eg: input handling: case event of: "keyUp:")
while main.nim goes into project folder - as it starts the show (app.run).

structure extended with files:
    ├── demo1       # Demo application
     |    └── demo.nim
     |  
    ├── simpleapp   # Application template
     |    └── main.nim
     |  
    └── multithreading
          └── channels  
           |    └── demo.nim
          └── threadpool
                └── demo.nim

src/nimble.nim
nais314: done.

src/nimble.nim
nais314: done.

src/nimblepkg/options.nim
nais314: done.

src/nimble.nim
dom96: Might be easier to create a separate PR for this. We can iterate faster that way.

dom96: You will need this as well to test this PR properly. I'm not sure how you're testing it currently as I don't see how the `examples` can be installed right now.

nais314: yes, examples/ lives in ~/.nimble/pkgs/pkg, just like tests/.
the idea is, that the package maintainer should/could add it, like tests/.
wether they using it or not, is up for the creator.

the main difference from eg tasks is, that i don't have to be in the packages directory,
i call nimble examples from my source dir. (so i have to lookup for pkgDir)

explanation:
i have a project,, the problem i face is that: to get a snippet from the pkgdir/examples/
i have to get the pkgdir, wich is not so easy,, except, if i hack nimble itself - and here we go ;)
...so in the end, i could publish, and write: 
"create a dir test, or whatever. type: nimble example myproj helloworld, then type nim c -r helloworld"

Yess, it is for convenience...  ;)
(((in defense, if needed: if a kid moves the example form gamengineX via drag and drop, then the kid will cry next time)))

((Other idea was, to extend tasks: nimble run pkgname taskname
wich would get currentDir and pkgDir.))

thank you for your time!

nais314: > You will need this as well to test this PR properly. I'm not sure how you're testing it currently as I don't see how the `examples` can be installed right now.

All right, i will think about how to write tests for it.
As examples/ should be added by the pkg writer, same way as tests/ - i sleep on it a little.
Thank again!

Araq:
@nais314 Why close it? It seems to be a good idea...

nais314:
Thank you!
Well, i jumped into it, short sighted, only looking at code parts i needed.
Now some tests are running fine, but looks like i violate the parser..huhh...
It will take more time. 
;)

nais314:
Thank you!
Well, i jumped into it, short sighted, only looking at code parts i needed.
Now some tests are running fine, but looks like i violate the parser..huhh...
It will take more time. 
;)

PR #645, opened 2019-05-02
Fix nocompiler for recursive nimble calls

genotrance:
Related to #635 and https://github.com/nim-lang/Nim/issues/11165.


dom96:
Test would be nice for this.

PR #646, opened 2019-05-03
Test case for #645 - recursive

genotrance:


tests/tester.nim
```diff
@@ -82,6 +82,10 @@ test "caching works":
     check output.contains("0.2.0")
     writeFile(nfile, readFile(nfile).replace("0.2.0", "0.1.0"))
 
+test "recursion works":
```
dom96: A little vague, please make these test names as explicit as possible. I'll merge this but for reference, this should mention "tasks" at the very least. My preferred name would be: "tasks can be called recursively"

Issue #80, opened 2014-12-30
nimble needs to support system-wide installations


philip-wernersbach:
(See http://forum.nim-lang.org/t/656 and http://github.com/Homebrew/homebrew/pull/34693)

I'd like to get nimble into some package managers alongside nim. In order to do this, nimble needs to support being installed to arbitrary system-wide locations, and needs to coexist with the nimble installation in a user's home directory. My solution to this is [nimble-wrapper](http://github.com/philip-wernersbach/nimble-wrapper), but there's probably a better way to do this.

philip-wernersbach:
Currently the only way to install to arbitrary locations is to modify the `nimbleDir` variable in `$AppDir/nimble/nimble.ini`. This is a no-go because packages can't install things to the user's home directory, so packages can't modify `$AppDir/nimble/nimble.ini` to set the correct location. Even if we could modify `$AppDir/nimble/nimble.ini`, doing so would take the installation in a user's home directory out of the equation.

For comparison, ruby's solution to this is that its library loading code supports multiple paths and specifies both system-wide paths and a path in the user's home directory by default. Ruby then iterates through the paths and uses the first one that contains the library to be loaded.

mindplay-dk:
I'm not an expert on this subject, but FWIW... I hated the way Ruby's package manager worked, where dependencies were installed and shared between projects. In contrast, I love how Composer installs dependencies to your project's own vendor folder, avoiding the clutter and potential conflicts associated with sharing dependencies between projects.

Because Nim links everything statically (and produces stand-alone executables by default) sharing dependent packages between projects probably doesn't even make any sense to begin with, so maybe this concern is irrelevant.

I also love how Node comes with NPM already installed out of the box - this is the thing I was requesting, primarily. If we have a working de-facto standard package manager for Nim, it should be available right away, encouraging people to create and consume packages and building an ecosystem :-)

philip-wernersbach:
Now that I think about it, I think @mindplay-dk's suggestion to package nimble with nim makes a lot of sense. I originally envisioned packaging nimble separately from nim, but nimble is such an essential part of the nim tooling that I don't really have "separation of concerns" issues with packaging them together. I also think that not having to install nimble separately would be a benefit to end users. This would also make my life easier when I start making Debian and Fedora packages for nim and nimble, as I would have to maintain half as many packages.

[...]

PR #656, opened 2019-05-25
Choose USER-specific tmpdir

cdunn2001:
re: #80

The question is what to do on Windows. Here, we check for `$USER` in the environment, and use it iff set.

The other question is how to find `/tmp`. It's more common to check `$TMPDIR` on Linux and OSX, and `getTempDir()` is discouraged in Nim.

This will use `$TMPDIR/$USER/nimble-$PID/nimblecache` if possible, and `getTempDir()/nimble-$PID/nimblecache` otherwise. That way current behavior can (almost) be maintained on systems which lack USER or TMPDIR, but people have ways around this when needed.

### DIFF
--- a/src/nimblepkg/nimscriptsupport.nim
+++ b/src/nimblepkg/nimscriptsupport.nim
@@ -15,7 +15,7 @@ from compiler/scriptconfig import setupVM
 from compiler/astalgo import strTableGet
 import compiler/options as compiler_options
 
-import common, version, options, packageinfo, cli
+import common, version, options, packageinfo, cli, tools
 import os, strutils, strtabs, tables, times, osproc, sets, pegs
 
 when not declared(resetAllModulesHard):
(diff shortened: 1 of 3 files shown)


src/nimblepkg/nimscriptsupport.nim
```diff
@@ -382,7 +382,7 @@ proc execScript(scriptName: string, flags: Flags, options: Options): PSym =
 
   # Ensure that "nimblepkg/nimscriptapi" is in the PATH.
   block:
-    let t = getTempDir() / "nimblecache"
+    let t = getNimbleTempDir() / "nimblecache"
```
dom96: This defeats the purpose of the cache, `getNimbleTempDir` will return a different directory for each Nimble invokation...

pb-cdunn: Oh! Duh. Ok.

pb-cdunn: How about now?

dom96:
I would change this to:

* Use the ``result`` var
* append the $USER to the result of ``getTempDir`` too

But this will do.

PR #658, opened 2019-05-29
Add backend selection to nimble init

3n-ki:
This adds a menu to select the backend within the `nimble init` subcommand.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -802,6 +802,14 @@ This should ideally be a valid SPDX identifier. See https://spdx.org/licenses/.
 Please specify a valid SPDX identifier.""",
       "MIT"
     )
+  var pkgBackend = options.promptList(
+    """Package Backend?
+c    - Compile using C backend.
+cpp  - Compile using C++ backend.
+objc - Compile using Objective-C backend.
+js   - Compile using JavaScript backend.""",
+    ["c", "cpp", "objc", "js"]
+  )
 
   # Ask for Nim dependency
   let nimDepDef = getNimrodVersion()
@@ -816,6 +824,7 @@ Please specify a valid SPDX identifier.""",
       pkgAuthor,
       pkgDesc,
       pkgLicense,
+      pkgBackend,
       pkgSrcDir,
       pkgNimDep,
       pkgType
--- a/src/nimblepkg/init.nim
+++ b/src/nimblepkg/init.nim
@@ -9,6 +9,7 @@ type
     pkgAuthor: string
     pkgDesc: string
     pkgLicense: string
+    pkgBackend: string
     pkgSrcDir: string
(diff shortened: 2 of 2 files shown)


src/nimble.nim
```diff
@@ -802,6 +802,15 @@ This should ideally be a valid SPDX identifier. See https://spdx.org/licenses/.
 Please specify a valid SPDX identifier.""",
       "MIT"
     )
+  var pkgBackend = options.promptList(
+    """Package Backend?
+c    - Compile to binary.
+cc   - Transpile to C.
```
dom96: Change these to "Compile using the C backend" (transpilation is a dirty word)

Also, remove the first option and rename ``cc`` to just `c`.

src/nimblepkg/init.nim
```diff
@@ -155,6 +156,7 @@ version       = $#
 author        = $#
 description   = $#
 license       = $#
+backend       = $#
```
dom96: Would be great if you could only add this when a backend that isn't `c` is selected.

src/nimble.nim
3n-ki: Will do! Mind if I ask why it's a dirty word? :P

src/nimblepkg/init.nim
3n-ki: Sure. How should I do that, use an `if...else` statement with two calls to `writeFile`? I'm not that well-versed in Nim yet, but i feel like it would have a cleaner way to do that.

dom96: I would create a variable to store the whole `"backend = $#"` string (set it to an empty string if the default backend is selected), then just add it to the final string at the end.

src/nimble.nim
dom96: Does gcc transpile to asm? :)

3n-ki: I guess not :)

src/nimblepkg/init.nim
3n-ki: Will do. I should have it done in a minute or two.

src/nimble.nim
GULPF: Since `c`/`cc` are semantically different (one means "use default backend" and the other "use the c backend") it might make sense to offer both. The default backend could theoretically be changed to `cpp` in the future.

3n-ki: Whoops, resolved this without seeing your comment. That's a good point, maybe we should consider that.

dom96:
Looks good, thank you :)

src/nimble.nim
dom96: We can solve that when the time comes :)

3n-ki:
Awesome, thanks! Always love to help out with anything Nim-related :)

PR #659, opened 2019-05-31
Fix outfile issue when binary exists

genotrance:
Issue reported [here](https://github.com/nim-lang/nimble/pull/635#issuecomment-497250610).

### DIFF
--- a/src/nimblepkg/nimscriptapi.nim
+++ b/src/nimblepkg/nimscriptapi.nim
@@ -35,6 +35,7 @@ var
   success = false
   retVal = true
   projectFile = ""
+  outFile = ""
 
 proc requires*(deps: varargs[string]) =
   ## Call this to set the list of requirements of your Nimble
@@ -42,13 +43,18 @@ proc requires*(deps: varargs[string]) =
   for d in deps: requiresData.add(d)
 
 proc getParams() =
+  # Called by nimscriptwrapper.nim:execNimscript()
+  #   nim e --flags /full/path/to/file.nims /full/path/to/file.out action
   for i in 2 .. paramCount():
     let
       param = paramStr(i)
-    if param.fileExists():
-      projectFile = param
-    elif param[0] != '-':
-      commandLineParams.add paramStr(i).normalize
+    if param[0] != '-':
+      if projectFile.len == 0:
+        projectFile = param
+      elif outFile.len == 0:
+        outFile = param
+      else:
+        commandLineParams.add param.normalize
 
 proc getCommand*(): string =
   return command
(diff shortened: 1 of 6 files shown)


src/nimblepkg/nimscriptapi.nim
```diff
@@ -35,20 +35,26 @@ var
...
-      projectFile = param
-    elif param[0] != '-':
-      commandLineParams.add paramStr(i).normalize
+    if param[0] != '-':
+      if projectFile.len == 0:
+        projectFile = param
+      elif outFile.len == 0:
+        outFile = param
+      else:
+        commandLineParams.add param.normalize
```
dom96: Is this code only parsing arguments that Nimble itself passes to `nim e`? It seems a little fragile to me, and if we control what gets passed to this anyway, can't we just serialise it into something concrete that is easily parsed?

src/nimblepkg/nimscriptwrapper.nim
```diff
@@ -42,12 +43,12 @@ proc execNimscript(nimsFile, projectDir, actionName: string, options: Options,
...
-      " " & nimsFileCopied.quoteShell & " " & actionName).strip()
+      " " & nimsFileCopied.quoteShell & " " & outFile.quoteShell & " " & actionName).strip()
+
+  displayDebug("Executing " & cmd)
 
   if live:
     result.exitCode = execCmd(cmd)
-    let
-      outFile = nimsFileCopied & ".out"
     if outFile.fileExists():
```
dom96: Do we still need this? The outFile is now in /tmp so it should be removed by the OS.

dom96:
Can confirm that this fixes the problems I was seeing.

src/nimblepkg/nimscriptapi.nim
genotrance: `nimscriptwrapper.nim:execNimscript()` calls `nim e` and the params are fixed in order. We don't need to tweak this beyond a point since it is only 3 params and we don't need to add yet another temp file into the mix.

src/nimblepkg/nimscriptwrapper.nim
genotrance: No harm deleting it either.

dom96:
Just ran into this:

Screen Shot 2019-06-01 at 00 06 41

Seems to have happened because I ran two `nimble` tasks in parallel.

genotrance:
I've made the .nims file unique using the PID so it should now allow running multiple calls in parallel. `nim e` gives a very random error when the nims file is not found.

Please confirm that this works for you.

dom96:
Works, thanks. I'll create a new release for this.

dom96:
Actually I'll do it tonight/tomorrow. Would be nice for @andreaferretti to confirm (please let us know if your `test` tasks work).

andreaferretti:
@dom96 @genotrance Works for me, thank you!

Issue #655, opened 2019-05-11
Nimble master has an exit code of zero, on Windows, on task failure


stefantalpalaru:
```text
Finished run: 51/53 tests successful
Error: execution of an external program failed: 'C:\Users\user\Desktop\status\nim-chronicles\tests\testrunner.exe .'
stack trace: (most recent call last)
..\..\..\AppData\Local\Temp\nimblecache\nimscriptapi.nim(157)
chronicles_4906363801196049117.nims(19) testTask
..\nimbus\vendor\Nim\lib\system\nimscript.nim(237) exec
..\nimbus\vendor\Nim\lib\system\nimscript.nim(237, 7) Error: unhandled exception: FAILED: nim c -r testrunner .

C:\Users\user\Desktop\status\nim-chronicles>echo %errorlevel%
0

C:\Users\user\Desktop\status\nim-chronicles>nimble --version
nimble v0.9.0 compiled at 2019-05-11 17:54:01
git hash: f17eaef7957da23aab6e375400b37227394a632a
```

PR #661, opened 2019-06-01
Fix issue #655

genotrance:
If child `nim e` process crashes, no output json is returned. In release mode, nimble looks at the output to verify if it should exit with non-zero value. Since it is blank, it doesn't.

I have updated the code to return an error string instead which results in the correct behavior.

### DIFF
--- a/src/nimblepkg/nimscriptwrapper.nim
+++ b/src/nimblepkg/nimscriptwrapper.nim
@@ -119,7 +119,13 @@ proc execScript(scriptName, actionName: string, options: Options):
     (output, exitCode) = execNimscript(nimsFile, scriptName.parentDir(), actionName, options)
 
   if exitCode != 0:
-    raise newException(NimbleError, output)
+    let
+      errOut =
+        if output.len != 0:
+          output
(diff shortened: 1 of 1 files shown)


src/nimblepkg/nimscriptwrapper.nim
```diff
@@ -119,7 +119,13 @@ proc execScript(scriptName, actionName: string, options: Options):
     (output, exitCode) = execNimscript(nimsFile, scriptName.parentDir(), actionName, options)
 
   if exitCode != 0:
-    raise newException(NimbleError, output)
+    let
+      errOut =
+        if output.len != 0:
+          output
+        else:
+          "Exception raised during nimble script execution"
```
dom96: IMO this would be easier to read:

```nim
var errMsg = output
if errMsg.len == 0:
  errMsg = "Exception raised during Nimble script execution"
```

dom96: Alternatively, even just removing the extra indentation after the `let` would be better.

PR #668, opened 2019-06-11
Add 0.20.0 to test matrix, latest nightlies

genotrance:


dom96:
Let's remove 0.19.6 once 0.20.0 is adopted: probably in a month or so.

PR #669, opened 2019-06-12
Add example dependency with git commit [docs]

hiteshjasani:
Just recently realized I could use a git url and commit hash in dependencies.  Thought others might benefit from seeing the example.


dom96:
Thanks! I modified it a little so that it's not encouraging users to use the git url for jester.

hiteshjasani:
Ah, good point!

PR #670, opened 2019-06-21
fixed #581 nimble init does not overwrite existing files with its templates anymore

### LINKED ISSUE #581: nimble init overwrites existing file src/myapp.nim when run in myapp/
Learned the hard way that running `nimble init` (v0.9.0) inside a `myapp/` dir will overwrite `src/myapp.nim` (at least for a binary package):

```
/tmp $ mkdir myapp

/tmp $ cd myapp

/tmp/myapp $ mkdir src

/tmp/myapp $ echo "# test" > src/myapp.nim

/tmp/myapp $ cat src/myapp.nim
# test

/tmp/myapp $ nimble init
      Info: Package initialisation requires info which could not be inferred.
        ... Default values are shown in square brackets, press
        ... enter to use them.
      Using "myapp" for new package name
      Using "myuser" for new package author
      Using "src" for new package source directory
    Prompt: Package type?
        ... Library - provides functionality for other packages.
        ... Binary  - produces an executable for the end-user.
        ... Hybrid  - combination of library and binary
        ... For more information see https://goo.gl/cm2RX5 [library/binary/hybrid]
binary
    Answer:     Prompt: Initial version of package? [0.1.0]

    Answer:     Prompt: Package description? [A new awesome nimble package]

    Answer:     Prompt: Package License?
        ... This should ideally be a valid SPDX identifier. See https://spdx.org/licenses/.
        ...  [MIT/GPL-2.0/Apache-2.0/ISC/GPL-3.0/BSD-3-Clause/LGPL-2.1/LGPL-3.0/EPL-2.0/Proprietary/Other]
MIT
    Answer:     Prompt: Lowest supported Nim version? [0.19.0]

    Answer:    Success: Package myapp created successfully

/tmp/myapp $ cat src/myapp.nim
# This is just an example to get you started. A typical binary package
# uses this file as the main entry point of the application.

when isMainModule:
  echo("Hello, World!")
```

liquid600pgm:


liquid600pgm:
@dom96 as you said on the IRC, using all the defaults is fine, so I did not change anything.

PR #677, opened 2019-07-16
Update submodules at checkout time to fix #675

### LINKED ISSUE #675: Nimbe pulls incorrect submodule version from a non master branch
Due to some version changes in the library I'm binding it is required to have different branches with different submodule versions from the `master` branch.

`1.0` branch: https://github.com/nimgl/nimgl/tree/1.0/src/nimgl/private
`master` branch: https://github.com/nimgl/nimgl/tree/master/src/nimgl/private

They both target different cimgui commit hashes, the issue comes when a user wants to install the branch `1.0` with `nimble install nimgl@#1.0`, it correctly downloads all files that are in the repo but fails at correctly updating the submodule, instead it just keeps the version that is referenced in the `master` branch which causes a SEGFAULT when trying to reference incorrect data: https://github.com/nimgl/nimgl/issues/25.
My naive idea of the error would be that doing the same with just the git cli actually requires an extra command: `git pull --recurse-submodules` after checkout the desired branch, which nimble lacks. 

This issue is not happening in a "mirror" repo that is kept as an independent package: https://github.com/nimgl/imgui it correctly downloads the desired submodule version.

lmariscal:
Doing the same as [download.nim#L38](https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/download.nim#L38) it updates the submodules recursively at the moment of checkout. It fixes https://github.com/nim-lang/nimble/issues/675.

Avoiding to use the `--init` argument since in [SO](https://stackoverflow.com/questions/10168449/git-update-submodules-recursively) they claim it can conflict with some older versions of git with already initialized submodules. Since nimble already clones with `--recursive` this shouldn't be a problem.

### DIFF
--- a/src/nimblepkg/download.nim
+++ b/src/nimblepkg/download.nim
@@ -24,6 +24,7 @@ proc doCheckout(meth: DownloadMethod, downloadDir, branch: string) =
       # clone has happened. Like in the case of git on Windows where it
       # messes up the damn line endings.
       doCmd("git checkout --force " & branch)
+      doCmd("git submodule update --recursive")
   of DownloadMethod.hg:
     cd downloadDir:
       doCmd("hg checkout " & branch)


dom96:
Awesome, thanks!

PR #682, opened 2019-07-26
Implement ability to pass compiler arguments with `nimble install -- …`

SolitudeSF:
Add `passNim/p` flag to `nimble install` for passing compiler flags. `nimble install --passNim:-d:danger`.

#599

### DIFF
--- a/src/nimblepkg/options.nim
+++ b/src/nimblepkg/options.nim
@@ -41,6 +41,7 @@ type
     of actionInstall, actionPath, actionUninstall, actionDevelop:
       packages*: seq[PkgTuple] # Optional only for actionInstall
                                # and actionDevelop.
+      passNimFlags*: seq[string]
     of actionSearch:
       search*: seq[string] # Search string.
     of actionInit, actionDump:
@@ -61,6 +62,7 @@ Usage: nimble COMMAND [opts]
 Commands:
   install      [pkgname, ...]     Installs a list of packages.
                [-d, --depsOnly]   Install only dependencies.
+               [-p, --passNim]    Forward specified flag to compiler.
   develop      [pkgname, ...]     Clones a list of packages for development.
                                   Symlinks the cloned packages or any package
                                   in the current working directory.
@@ -175,6 +177,7 @@ proc initAction*(options: var Options, key: string) =
   case options.action.typ
   of actionInstall, actionPath, actionDevelop, actionUninstall:
     options.action.packages = @[]
+    options.action.passNimFlags = @[]
   of actionCompile, actionDoc, actionBuild:
     options.action.compileOptions = @[]
     options.action.file = ""
@@ -303,6 +306,8 @@ proc parseFlag*(flag, val: string, result: var Options, kind = cmdLongOption) =
       case f
       of "depsonly", "d":
         result.depsOnly = true
+      of "passnim", "p":
+        result.action.passNimFlags.add(val)
       else:
         wasFlagHandled = false
     of actionUninstall:
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -213,13 +213,13 @@ proc processDeps(pkginfo: PackageInfo, options: Options): seq[PackageInfo] =
   for i in reverseDeps:
     addRevDep(options.nimbleData, i, pkginfo)
 
-proc buildFromDir(pkgInfo: PackageInfo, paths: seq[string],
-                  args: var seq[string]) =
+proc buildFromDir(pkgInfo: PackageInfo, paths, args: seq[string]) =
   ## Builds a package as specified by ``pkgInfo``.
   if pkgInfo.bin.len == 0:
     raise newException(NimbleError,
         "Nothing to build. Did you specify a module to build using the" &
         " `bin` key in your .nimble file?")
+  var args = args
   let realDir = pkgInfo.getRealDir()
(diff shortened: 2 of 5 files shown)


c-blake:
I think a feature like this would be widely appreciated!

src/nimble.nim.cfg
```diff
@@ -4,3 +4,4 @@
 --path:"./vendor/nim"
 -d:ssl
 -d:nimcore # Enable 'gorge' in Nim's VM. See https://github.com/nim-lang/Nim/issues/8096
+-d:nimOldCaseObjects
```
genotrance: Why is this needed now?

src/nimblepkg/options.nim
```diff
@@ -40,6 +40,7 @@ type
     of actionInstall, actionPath, actionUninstall, actionDevelop:
       packages*: seq[PkgTuple] # Optional only for actionInstall
                                # and actionDevelop.
+      installOptions*: seq[string]
```
genotrance: `installOptions` isn't too clear a name

SolitudeSF: well, i need some alternative to `compileOptions2`

src/nimble.nim.cfg
SolitudeSF: when i built nimble manually, without koch, it crashed without this flag.

genotrance:
Two general comments:
- Can this be done without the extra `--`? Are there any nimble flags which apply to nim as well that won't get passed through?
- What about `nimble build`?

SolitudeSF:
This can be done without `--`, nimble would just have to sacrifice short `-d` option for deps only.

I was under the impression that nimble build already did that

src/nimble.nim.cfg
genotrance: You are using a really old version of nimble. Need to get the latest codebase.

SolitudeSF: Ah, forgot to `pull`

src/nimblepkg/options.nim
SolitudeSF: would `installCompileOptions` be better?

c-blake: It sure is crystal clear (though a bit of a mouthful, but that's usually the trade-off).

genotrance: Why not something like `commandLineFlags` or something that suggest they come from the CLI.

src/nimblepkg/options.nim
```diff
@@ -60,7 +61,9 @@ Usage: nimble COMMAND [opts]
 
 Commands:
   install      [pkgname, ...]     Installs a list of packages.
-               [-d, --depsOnly]   Install only dependencies.
+               [-D, --depsOnly]   Install only dependencies.
```
genotrance: This will surely break someone. Cannot change existing flags.

I presume you are trying to enable `-d:defineToPassNim` but does `-d:xyz` still get caught in the nimble code? Isn't it just looking for `-d`?

src/nimblepkg/options.nim
SolitudeSF: well, i took whats already implemented for `build` command. should it be changed too or just this instance?

src/nimblepkg/options.nim
SolitudeSF: yep, it will. i can add workaround for that.

src/nimblepkg/options.nim
genotrance: I'd prefer that but @dom96 might want to leave it as is and consistent.

src/nimblepkg/options.nim
SolitudeSF: added workaround. now only `-d` is captured by nimble.

genotrance:
Test cases would be great.

Otherwise LGTM, let's wait from @dom96's review.

SolitudeSF:
whats the point of addidng `passNim` to `build` is it has no cli flags by itself? or is this just futureproofing?

dom96:
> or is this just futureproofing?

It is, although I'm willing to make ``nimble build`` an exception just like ``nimble c`` (they already allow ``-d:release`` IIRC).

SolitudeSF:
both `build` and `c` forward all flags to compiler

SolitudeSF:
i dont really understand what to do with custom task. i see the `arguments` and `flags` fields in `Action`, but i dont see it used.

dom96:
> i dont really understand what to do with custom tasks. i see the arguments and flags fields in Action, but i dont see them being used.

Yeah, that's a bit more complex, feel free to leave that for another time

SolitudeSF:
@genotrance, how would i go about testing this? this just uses already existing mechanism that just wasnt utilized fully.

src/nimblepkg/options.nim
dom96: This needs to have ``passNim`` in the name... rename it to ``passNimFlags``.

genotrance:
Well given you have added the ability to pass flags to the install step, you will need to create a test project that needs some -d:xyz to compile and install properly. Perhaps also a custom gc or other Nim flag that results in a special when statement being executed.

SolitudeSF:
Added test. If its correct then that should be it.

dom96:
Awesome, looks good, thank you!

PR #683, opened 2019-07-26
Fixes #504

### LINKED ISSUE #504: `nimble remove` skips some packages
I'm not sure what it's based on, but `nimble remove` sometimes skips one of the packages it's given:
```
[dmitry@dmitry nimfoo]$ nimble install -y nimfp@0.4.2 nimboost@0.5.4
Downloading https://github.com/vegansk/nimfp using git
   Warning: Package 'nimfp' has an incorrect structure. The top level of the package source directory should contain at most one module, named 'nimfp.nim', but a file named 'fp.nim' was found. This will be an error in the future.
      Hint: If this is the primary source file in the package, rename it to 'nimfp.nim'. If it's a source file required by the main module, or if it is one of several modules exposed by 'nimfp', then move it into a 'nimfp/' subdirectory. If it's a test file or otherwise not required to build the the package 'nimfp.nim', prevent its installation by adding `skipFiles = @["fp.nim"]` to the .nimble file. See https://github.com/nim-lang/nimble#libraries for more info.
  Verifying dependencies for nimfp@0.4.2
 Installing nimboost@>= 0.5.3
Downloading https://github.com/vegansk/nimboost using git
   Warning: Package 'nimboost' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'nimboost', but file 'richstring.nim' is in a directory named 'boost' instead. This will be an error in the future.
      Hint: If 'boost' contains source files for building 'nimboost', rename it to 'nimboost'. Otherwise, prevent its installation by adding `skipDirs = @["boost"]` to the .nimble file.
  Verifying dependencies for nimboost@0.5.4
      Info: Dependency on patty@>= 0.3.1 already satisfied
  Verifying dependencies for patty@0.3.1
 Installing nimboost@0.5.4
   Success: nimboost installed successfully.
      Info: Dependency on classy@>= 0.0.2 already satisfied
  Verifying dependencies for classy@0.0.2
 Installing nimfp@0.4.2
   Success: nimfp installed successfully.
Downloading https://github.com/vegansk/nimboost using git
  Verifying dependencies for nimboost@0.5.4
      Info: Dependency on patty@>= 0.3.1 already satisfied
  Verifying dependencies for patty@0.3.1
 Installing nimboost@0.5.4
    Prompt: nimboost@0.5.4 already exists. Overwrite? -> [forced yes]
   Success: nimboost installed successfully.
[dmitry@dmitry nimfoo]$ nimble remove -y nimfp nimboost
    Looking for nimfp (any version)
   Checking reverse dependencies
    Looking for nimboost (any version)
   Checking reverse dependencies
    Prompt: The following packages will be removed:
        ...   nimfp (0.4.2)
        ... Do you wish to continue? -> [forced yes]
    Removed nimfp (0.4.2)
```
note that `nimboost`'s reverse dependencies are checked, but it is not removed.

My guess is that this probably based on dependencies (`nimfp` depends on `nimboost`), but in any case `nimble remove` shouldn't silently skip the specified package.

genotrance:
Fixes #504 - better behavior for multi-package removal

- Better error message when nimble cannot uninstall something due to a reverse dependency

> nimble remove nimboost nimfp

image

`nimboost` cannot be removed since `nimfp` depends on it but `nimfp` hasn't been processed yet.

- Uninstall if packages specified in the right order - i.e. reverse dependency was already processed for removal

> nimble remove nimfp nimboost

image

`nimboost` can be removed since its reverse dependency `nimfp` was already processed and marked for removal.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -840,7 +840,8 @@ proc uninstall(options: Options) =
     raise newException(NimbleError,
         "Please specify the package(s) to uninstall.")
 
-  var pkgsToDelete: seq[PackageInfo] = @[]
+  var pkgsToDelete: HashSet[PackageInfo]
+  pkgsToDelete.init()
   # Do some verification.
   for pkgTup in options.action.packages:
     display("Looking", "for $1 ($2)" % [pkgTup.name, $pkgTup.ver],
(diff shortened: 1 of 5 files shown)


genotrance:
@dom96 - PTAL

tests/tester.nim
```diff
@@ -827,3 +826,17 @@ test "init does not overwrite existing files (#581)":
     check execNimbleYes("init").exitCode == QuitSuccess
     check readFile("src/issue581.nim") == Src
   removeDir("issue581")
+
+test "remove skips packages with revDeps (#504)":
+  check execNimble("install", "nimboost", "nimfp", "-y").exitCode == QuitSuccess
```
dom96: It's great that you've added a test here, but this looks flaky. At least please specify a commit hash so that it's less likely that an installation will fail.

src/nimble.nim
```diff
@@ -851,14 +851,20 @@ proc uninstall(options: Options) =
...
-    var errors: seq[string] = @[]
     for pkg in pkgList:
       # Check whether any packages depend on the ones the user is trying to
       # uninstall.
       if options.uninstallRevDeps:
         getAllRevDeps(options, pkg, pkgsToDelete)
       else:
-        let revDeps = getRevDeps(options, pkg)
+        let
+          # Pkgs already processed for removal
```
dom96: The word "processed" here is misleading, it implies that these packages have already been deleted but this isn't the case. Rephrase this to something like this: "Names of packages already scheduled for removal"

(Please use full words in comments instead of abbreviations, it makes it harder to read, same for `revDeps` below).

src/nimble.nim
```diff
@@ -851,14 +851,20 @@ proc uninstall(options: Options) =
...
+          namesOfPkgsToDelete = pkgsToDelete.mapIt(it.name)
```
dom96: This is a good start, but it doesn't take package versions into account at all. Users can specify version requirements when removing packages: ``nimble remove foo@0.1.0``.

I think what might be a good approach to this is to refactor ``pkgsToDelete`` into a ``HashSet[PackageInfo]``, you'll need to implement a ``hash()`` function for ``PackageInfo`` and ensure the hash is based on the package name + version. If all goes well you'll be able to write ``if len(revDeps * pkgsToDelete) > 0:`` below instead of ``revDeps.len > 0`` and this mapping/filtering logic.

tests/tester.nim
genotrance: I've linked to specific versions.

src/nimble.nim
genotrance: I've updated per our discussion in IRC. Had to be `revDeps - pkgsToDelete`.

dom96:
Looks good, thanks!

Issue #410, opened 2017-09-28
Foreign characters input with nimble init are escaped


mjoud:
`nimble init` will escape "foreign characters" since `strutils.escape` is used. This leads to a surprising `.nimble` file.

```
$ nimble init test
      Info: In order to initialise a new Nimble package, I will need to ask you
        ... some questions. Default values are shown in square brackets, press
        ... enter to use them.
    Prompt: Initial version of package? [0.1.0]
    Answer: 
    Prompt: Your name? [My name]
    Answer: Mañana
    Prompt: Package description?
    Answer: Blött vått slätt    
    Prompt: Package license? [MIT]
    Answer: 
    Prompt: Lowest supported Nim version? [0.17.2]
    Answer: 
   Success: Nimble file created successfully
       Info Hint: used config file '/etc/nim.cfg' [Conf]

$ cat test.nimble 
# Package

version       = "0.1.0"
author        = "Ma\xC3\xB1ana"
description   = "Bl\xC3\xB6tt v\xC3\xA5tt sl\xC3\xA4tt"
license       = "MIT"

# Dependencies

requires "nim >= 0.17.2"


```

DestyNova:
This is affecting me too... I have to manually fix my name in the .nimble file after every `nimble init`. Can we just not call `strutils.escape` when populating those fields? I'm not sure what it's trying to guard against.

dom96:
@DestyNova it's making sure that any non-printable characters are escaped, but you're right, this is unlikely to happen for names. Happy to accept a PR that remove ``strutils.escape``.

DestyNova:
PR opened :smiley:

DestyNova:
(Note: I also removed escaping of the package description because I can easily imagine non-English characters being used there.)

genotrance:
This is fixed.

PR #684, opened 2019-07-27
Remove escaping of author name and package description in `nimble init`

DestyNova:
...since it replaces characters used in many names and languages with escape sequences.

Slightly refactor the code for determining author name, to make it easier to add other version control systems in future (right now it's just git and hg).

Also, add some binary test artifacts to .gitignore so they don't accidentally get committed in future.

Fixes issue #410.

### DIFF
--- a/src/nimblepkg/init.nim
+++ b/src/nimblepkg/init.nim
@@ -175,7 +175,7 @@ $#
 
 requires "nim >= $#"
 """ % [
-      info.pkgVersion.escape(), info.pkgAuthor.escape(), info.pkgDesc.escape(),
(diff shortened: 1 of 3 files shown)


src/nimblepkg/init.nim
```diff
@@ -175,7 +175,7 @@ $#
 
 requires "nim >= $#"
 """ % [
-      info.pkgVersion.escape(), info.pkgAuthor.escape(), info.pkgDesc.escape(),
+      info.pkgVersion.escape(), info.pkgAuthor, info.pkgDesc.escape(),
```
dom96: package description still escaped, no?

DestyNova: Oops!

dom96:
Thanks!

PR #686, opened 2019-07-30
Fix #633 - pass CLI to tasks

### LINKED ISSUE #633: Passing command line arguments to a task
It would be nice if nimble tasks could be parameterized by the user via the command line. For example: `nimble myTask --verbose` and the task could handle the `--verbose` switch.

Not sure if this is already possible, but I couldn't figure out how.

genotrance:
Args and flags are now forwarded to the tasks in this PR but considering nimble now uses `nim e`, we have to pass a bunch of other flags to the compiler which results in a long list of stuff that task writers have to skip.

From the original issue:
```
nimble dmdm --abc
  Executing task dmdm in /home/binhong/wings/wings.nimble
7
/home/binhong/.choosenim/toolchains/nim-0.20.2/bin/nim
e
--hints:off
--verbosity:0
-p:/tmp/nimblecache
/home/binhong/wings/wings_21085.nims
/tmp/nimble_21085.out
dmdm
```

Any suggestions to simplify or clean this up will be helpful.

### DIFF
--- a/src/nimblepkg/nimscriptwrapper.nim
+++ b/src/nimblepkg/nimscriptwrapper.nim
@@ -40,9 +40,17 @@ proc execNimscript(nimsFile, projectDir, actionName: string, options: Options):
     if not isScriptResultCopied:
       nimsFileCopied.removeFile()
 
-  let
+  var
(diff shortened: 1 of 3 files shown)


dom96:
This can do for now, I don't have better ideas. Can we get a test for it too though?

PR #688, opened 2019-08-02
Fix #567: Display package versions in sorted order

### LINKED ISSUE #567: package versions are not ordered
```
$ nimble search kotlin --ver
pykot:
  url:         https://github.com/jabbalaci/nimpykot (git)
  tags:        library, python, kotlin
  description: Porting some Python / Kotlin features to Nim
  license:     MIT
  website:     https://github.com/jabbalaci/nimpykot
  versions:    v0.1.2, v0.1.3, v0.2.0, v0.2.1, v0.1.0, v0.1.1
```

Versions should be ordered in my opinion. Maybe a descending order would be better, thus the newest version would be at the beginning.

hiteshjasani:
Package versions are parsed into semantic versions (if possible) and sorted in descending order.  The most recent version is shown first.

Here are some before and after examples.

**Before**
```
genieos:
versions:    v9.4.0-tiffany, v9.2.0-sunny, v9.0.0-taeyeon, v9.4.2-hyoyeon, v9.
0.1-jessica

libtcod_nim:
versions:    v0.9, v0.95, v0.97, v0.98, v0.99, v0.98.1, v0.96, v1.112.0

nimfp:
versions:    v0.4.1, v0.0.5, v0.0.7, v0.3.0, v0.2.4, v0.0.1, v0.4.3, v0.3.7, v0.0.6,
v0.2.2, v0.3.2, v0.0.4, v0.3.3, v0.3.6, v0.2.0, v0.4.4, v0.4.0, v0.0.3, v0.2.1,
v0.0.2, v0.1.0, v0.0.9, v0.4.2, v0.2.3, v0.3.1, v0.3.4, v0.0.8, v0.3.5
```

**After**
```
genieos:
versions:    v9.4.2-hyoyeon, v9.4.0-tiffany, v9.2.0-sunny, v9.0.1-jessica, v9.
0.0-taeyeon

libtcod_nim:
versions:    v1.112.0, v0.99, v0.98.1, v0.98, v0.97, v0.96, v0.95, v0.9

nimfp:
versions:    v0.4.4, v0.4.3, v0.4.2, v0.4.1, v0.4.0, v0.3.7, v0.3.6, v0.3.5, v0.3.4,
v0.3.3, v0.3.2, v0.3.1, v0.3.0, v0.2.4, v0.2.3, v0.2.2, v0.2.1, v0.2.0, v0.1.0,
v0.0.9, v0.0.8, v0.0.7, v0.0.6, v0.0.5, v0.0.4, v0.0.3, v0.0.2, v0.0.1
```


src/nimblepkg/semver.nim
```diff
@@ -0,0 +1,73 @@
+import algorithm, sequtils, strutils
```
dom96: Nimble already has a version parsing module, could you re-use it instead of introducing this new one? (https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/version.nim)

readme.markdown
```diff
@@ -786,6 +786,23 @@ To summarise, the steps for release are:
 Once the new tag is in the remote repository, Nimble will be able to detect
 the new version.
 
+##### Git Version Tagging
```
dom96: I don't think this information is important enough to include in the readme. We should be making the Nimble docs as streamlined as possible.

src/nimblepkg/semver.nim
hiteshjasani: Just pushed a commit that should address this.

readme.markdown
hiteshjasani: There is an implicit relationship between the git tag label and the version info that nimble displays and uses.  It's valuable to let users know some info about it so they don't have to guess or dive into the code just to remember how they should be tagging their releases.  I'm still going to circle back to this and see if I can streamline the docs a bit more.

dom96: I would prefer if you instead emphasise that this is how users should be tagging in the above section. All the details about version sorting is unnecessary.

hiteshjasani:
I'm not sure what is causing the build failure.  Do you have any insight?

dom96:
It looks like you broke Nimble compilation on 0.19.6. We probably want to disable that now since 0.19.6 is fairly old. But for now can you attempt to fix it if it's not too much trouble? Alternatively you can disable this in travis.yml.

hiteshjasani:
I think all the issues have been addressed. Let me know if there is anything else.

dom96:
Nice, thanks!

PR #691, opened 2019-08-05
Extended info

pmetras:
I have been curating all the Nimble packages' web sites in order to add better package selection criteria (code has been updated in https://github.com/pmetras/packages, branch extended_info but not the packages that are in https://docs.google.com/spreadsheets/d/1HWy2YumMMcgEDHk34ACauuWR5TYDJTRUVQ6B-LuRXCs/edit?usp=sharing and for which I'll need to create a JSON file when curration is completed).

I've added new metadata fields to the `packages.json` file. In particular:
* `long_description`: a multilines description of the package.
* `categories`: broad categories that can be used to group the packages.
* `code_quality`, `doc_quality` and `project_quality`: subjective metrics about the code, documentation and project that are used to calculate a maturity metric that can be used to select packages.
* `logo`: the URL to the project logo (not used presently).
* `screenshots`: list of URLs to project screenshots (not used presently).

You can see more information about these fields in https://github.com/pmetras/packages/blob/extended_info/README.md.

New options were added to `search` and `list`:
* `--full` displays the new metadata when not null and search into long description and categories.
* `--mat=M` selects only packages where maturity level is superior to `M`, in a range 0.0 (dead package or not rated yet) to 4.0 (high quality community supported package).

`search`, `list` and `publish` actions have been changed to benefit from the new metadata fields when using the `--full` option, searching in the long description and categories and displaying them. 

The `publish` action hasn't been tested yet as I don't use a certificate. _**I'm looking for a tester...**_

Nimble still remains compatible with the previous version as these fields are optional.


pmetras:
Following discussions on [Nim forum](https://forum.nim-lang.org/t/5190), I had decided to drop the quantitative variables `..._quality` where we couldn't agree on an evaluation method and whose maintenance would be problematic. But I wanted to keep the extended documentation and categories into Nimble `packages.json`, as this can be very useful to find packages.

As an example, let say you want to find a web server written in Nim. `nimble search web` will return a list of packages including microasynchttpserver, mofuw or nimhttpd. Have you spotted that httpbeast is not part of the results, nor httputils or whip? You can find them with `nimble search --full web`.

I still have to adapt the code and the `packages.json` file though...

dom96:
That is what the tags are for, why have both categories and tags?

Extended documentation is already covered by the underlying package's git repo (see how nimble.directory shows it).

pmetras:
Well, tags are different from categories, at least in information management. Categories are fixed and commonly shared "tags" that are used to categorize (I don't know how to better explain it in English). Main use is to group similar results in a search. If I want to list all packages related to "Data Science" or "Education", I can do that with categories. They can also be used to browse packages in a graphical package browser.

Regarding the `extended_doc` field, the goal is not to reproduce the README file from Github but to provide a synopsis of the package in more than 1 line. And more importantly, at least for me, to be able to have a search function in nimble that is more broad/deep, like I demonstrated in my previous example. If all packages README files are centralized on nimble.directory and used to search through and nimble could access the search results, then the `extended_doc` field would be useless. I use nimble as a package search and discovery command line tool more than my web browser...

I understand that putting too much information in the `packages.json` file makes maintenance of the file more difficult. Perhaps plugging nimble into http://nimble.directory API would be a better route, at least for search functions.

pmetras:
So what has been decided? I won't do the changes to the code and more importantly to the `packages.json` data if they are considered useless.
If you don't want to add the changes into nimble, that's not a problem. I'll try to help improving http://nimble.directory/.

dom96:
Yeah, I would prefer to keep packages.json slim. I think this information can definitely go into nimble.directory or a separate additional JSON file that Nimble can optionally download.

genotrance:
What's the verdict on this one?

dom96:
I think we should close it. I don't think this information belongs in Nimble's packages.json.

Issue #680, opened 2019-07-20
Clear build warnings


bobeff:
I have desire to clear all build warnings form Nimble, but I'm not sure what to do with deprecation warnings. It seems that Nimble is still being tested against an older version of Nim - [0.19.6](https://github.com/nim-lang/nimble/blob/7c2b9f612e6a42609f28f8f51de4ab6949ced4a0/.travis.yml#L8). Half of the warnings are of this type:
```
Warning: Deprecated since v0.20, use 'initHashSet'; initSet is deprecated [Deprecated]
```
and fixing them will break backward compatibility with an older Nim. Is such compatibility still required?

The other half of the warnings are from the type:
```
Warning: imported and not used: 'sets' [UnusedImport]
```

and they are caused by the new feature for detecting unused imports. I suspect that it will be completely safe to remove them.

There are also some hints for dead code like:
```
Hint: 'getSpecificDir' is declared but not used [XDeclaredButNotUsed]
```

Can I remove such unused functions or they are left intentionally?

bobeff:
If the compatibility with an older Nim before version 0.20.x is still required, I can fix it this way:
```nim
when NimMajor == 0 and NimMinor < 20:
  import sets

  template initHashSet*[A](initialSize = 64): HashSet[A] =
    initSet[A](initialSize)

  template toHashSet*[A](keys: openArray[A]): HashSet[A] =
    toSet(keys)
```

cheatfate:
@bobeff it is better to use `compiles` e.g:
```nim
when compiles(initHashSet):
  # After 0.20.0 Nim's stdlib version
  result.handles = initHashSet[AsyncFD]()
else:
  # Pre 0.20.0 Nim's stdlib version
  result.handles = initSet[AsyncFD]()

when compiles(initHeapQueue):
  # After 0.20.0 Nim's stdlib version
  result.timers = initHeapQueue[TimerCallback]()
else:
  # Pre 0.20.0 Nim's stdlib version
  result.timers = newHeapQueue[TimerCallback]()
```

dom96:
You don't need ``compiles`` for this, ``declared`` is more direct and should work (https://nim-lang.org/docs/system.html#declared%2Cuntyped).

dom96:
> Can I remove such unused functions or they are left intentionally?

Sure, feel free to remove them. If it turns out to be important we'll flag it in PR reviews.

bobeff:
[...]

Issue #689, opened 2019-08-02
`nimble list --ver` fails on nim-locale package


hiteshjasani:
`nimble list --ver --verbose` fails partway through listing packages with the following message:

```
nim-locale:
  Alias for locale
     Error: Invalid download method:
```

bobeff:
It seems that for some reason `git ls-remote <some_url>` fails from time to time. I think that for the same reason [this build](https://travis-ci.org/nim-lang/nimble/builds/568431732) of my [pull request](https://github.com/nim-lang/nimble/pull/692) failed.

Is it possible GitHub to have some kind of bot protection which is being activated when there are too many request from a single IP address for a short amount of time, which happens when unit tests are being run and when `nimble list` command is being executed?

dom96:
The ``search`` command also supports ``--ver``, likely reuses these code paths

PR #692, opened 2019-08-06
Fix build warnings

bobeff:
In this pull request related to #680 Nim's build warnings while compiling Nimble are cleared. In a few separate commits are fixed different types of warnings:

* Unused imports are fixed in e8c7d5c
* Deprecation warnings are fixed in 3d6172e
* Unused symbols warnings are fixed in 7df2ef3
* Compilation of the nimscriptapi.nim file is fixed in 19f160e


bobeff:
@dom96  Did you see [this](#689) and [my comment](https://github.com/nim-lang/nimble/issues/689#issuecomment-518796837) on it? Can you rerun the unit tests manually from travis UI just for checking my assumption, please? It seems that there is such an [option](https://blog.travis-ci.com/2017-08-24-trigger-custom-build), but maybe I don't have permissions for it, because I didn't see it available in the UI.

dom96:
Restarted. Indeed, seems like there are issues accessing GitHub. That really sucks :(

genotrance:
That's a connection failure, not a rate limit.

bobeff:
I think that this pull request is already at its final version in 44fc449. Additionally to the original pull request:

* Unused file nimscriptsupport.nim is removed in 134b799.
* Unit test for checking that Nimble compiles without already removed warning types is added in e1b6719 and it is fixed to pass in 2f24cd5 and 500089f.

bobeff:
@dom96 What is special with 44aadd50cfa647a759610a15967960632bf597ce version of Nim, that Nimble is being tested against it? I found a compilation problem with the latest Nim and I'm wandering whether it will be more appropriate to test against `#head` instead?

dom96:
> What is special with nim-lang/Nim@44aadd5, that Nimble is being tested against it?

@bobeff there is nothing special. It's just a pinned commit. Might be worth to add a comment to make this clear in there.

src/nimblepkg/download.nim
```diff
@@ -29,19 +22,6 @@ proc doCheckout(meth: DownloadMethod, downloadDir, branch: string) =
...
-    doCheckout(meth, downloadDir, "")
-    cd downloadDir:
-      doCmd("git pull")
-      if existsFile(".gitmodules"):
-        doCmd("git submodule update")
-  of DownloadMethod.hg:
-    doCheckout(meth, downloadDir, "default")
-    cd downloadDir:
-      doCmd("hg pull")
-
```
dom96: While these are not used currently I think they're worth keeping, they might be useful soon.

src/nimblepkg/common.nim
```diff
@@ -64,3 +63,15 @@ when not defined(nimscript):
...
   nimbleVersion* = "0.10.2"
+
+when not declared(initHashSet):
+  import sets
+
+  template initHashSet*[A](initialSize = 64): HashSet[A] =
+    initSet[A](initialSize)
+
+when not declared(toHashSet):
+  import sets
```
dom96: This might fail if the above ``import`` is present too, have you checked whether that's the case?

bobeff:
@dom96 What about testing against `#head` instead? I don't see a good reason to test against arbitrary commits, but I just found two issues with the latest Nim.

genotrance:
While I prefer knowing asap that nimble is broken with devel, @dom96 prefers testing against a stable Nim. A comment will be useful. Please bump it up to the latest nightly that passed while you are at it.

src/nimblepkg/packageinstaller.nim
```diff
@@ -3,7 +3,13 @@
 import os, strutils, sets, json
 
 # Local imports
-import cli, common, options, tools
+import cli, options, tools
+
+when defined(windows):
+  import version
```
dom96: Why is this just imported for Windows?

tests/tester.nim
```diff
@@ -10,6 +10,7 @@ var rootDir = getCurrentDir().parentDir()
 var nimblePath = rootDir / "src" / addFileExt("nimble", ExeExt)
 var installDir = rootDir / "tests" / "nimbleDir"
 const path = "../src/nimble"
+const stringNotFound = -1
```
dom96: Nice :)

tests/tester.nim
```diff
@@ -859,4 +860,40 @@ test "do not install single dependency multiple times (#678)":
       check execNimble(["refresh"]).exitCode == QuitSuccess
       let (output, exitCode) = execNimble("install", "-y")
       check exitCode == QuitSuccess
-      check output.find("issue678_dependency_1@0.1.0 already exists") == -1
+      let index = output.find("issue678_dependency_1@0.1.0 already exists")
+      check index == stringNotFound
+
+test "compilation without warnings":
```
dom96: Not too sure I want this to be a test. Also, doesn't Nim offer a flag that turns warnings into errors?

dom96:
@genotrance no, I don't prefer testing against a stable Nim. I prefer testing against a known working version of Nim. The test suite is already flaky enough as it is, if the tester runs against a different Nim build each time it'll make it even more flaky.

Pinning commits like this is a common thing to do. Testing against a moving target is never a good idea.

bobeff:
>  I prefer testing against a known working version of Nim.

Maybe separate build badge for testing against stable and dev Nim versions can be used, but unfortunately Travis CI is still not supporting this travis-ci/issues#5035.

bobeff:
I added one more commit b4d8dbf with two problems with the latest Nim fixed.

src/nimblepkg/packageinstaller.nim
bobeff: Because only on Windows something from this module is being used. Namely the `NimbleError` exception which is being thrown below. See [here](https://github.com/nim-lang/nimble/blob/44fc44980e7473f7c3e5c93d42aca65c07d90899/src/nimblepkg/packageinstaller.nim#L52).

src/nimblepkg/common.nim
bobeff: I tested that there is not an error, nor a warning, when a used module is being imported multiple times.

src/nimblepkg/download.nim
bobeff: Do you have something specific in mind? I will return them if so, but my issue with them was that just recently, when I had added `local` download method, I was required to maintain their code also, despite that they are not being used. I'm expecting soon to work on adding `tarball` download method and such a maintenance will be required again.

genotrance:
> @genotrance no, I don't prefer testing against a stable Nim. I prefer testing against a known working version of Nim (that is very close to #devel). 

What we really need is choosenim support to install nightlies. Nightlies only get posted when the build succeeds so it is a known working build. I've been manually updating the pinned commit with nightlies, will be nice to have it automated.

dom96:
@genotrance Updating the commit manually every so often isn't so difficult. Even if we know that the build we use for #head built successfully that still doesn't guarantee that there isn't a regression that broke Nimble (and that's something that Nim repo tests for so I don't see why Nimble's CI should break too).

narimiran:
I was about to send my PR dealing with these, just to realize — it already exists!

@bobeff 
I quickly skimmed through your changes and I don't see that you've changed the stuff that was producing `Hint: simpleGetOrDefault(j, "success") --> 'getOrDefault(j, "success")' [Pattern]`.

If you want to add this too, here is one example of what needs to be changed:
```nim
# original
    result = j{"parent"}{"full_name"}.getStr() == "nim-lang/packages"

# new
    result = j.getOrDefault("parent").getOrDefault("full_name").getStr() == "nim-lang/packages"
```

src/nimblepkg/download.nim
dom96: yes, the plan is to make Nimble manage git repos instead of re-cloning packages at different commit hashes all the time.

bobeff:
> I quickly skimmed through your changes and I don't see that you've changed the stuff that was producing Hint: simpleGetOrDefault(j, "success") --> 'getOrDefault(j, "success")' [Pattern].

@narimiran As I understood this is not a warning, but just note about pattern expansion as result of term rewriting macro.

src/nimblepkg/download.nim
bobeff: Reverted in 057505b.

bobeff:
> can you rebase?

@dom96 Done. I think that it is ready for merge.

PR #697, opened 2019-09-03
Fix #689 - don't lookup versions for aliases

### LINKED ISSUE #689: `nimble list --ver` fails on nim-locale package

genotrance:
Both `nimble list --ver` and `nimble search xyz` code paths are fixed.


dom96:
Tests would be great if you're up for it

PR #698, opened 2019-09-04
Fix #640 - strip out blank spaces

### LINKED ISSUE #640: nimble dump wants an answer

genotrance:


dom96:
Test would be awesome if you're up for it :)

PR #702, opened 2019-09-16
Option to provide JSON output from search and list commands

ehmry:
JSON output is helpful for transforming Nimble metadata into other kinds of packaging.

Ref https://github.com/NixOS/nixpkgs/pull/67878

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -581,12 +581,20 @@ proc search(options: Options) =
   if needsRefresh(options):
     raise newException(NimbleError, "Please run nimble refresh.")
   let pkgList = getPackageList(options)
-  var found = false
+  var
+    found = false
+    jsonOutput =  newJArray()
   template onFound {.dirty.} =
-    echoPackage(pkg)
-    if pkg.alias.len == 0 and options.queryVersions:
-      echoPackageVersions(pkg)
-    echo(" ")
+    if options.jsonOutput:
+      let info = pkg.toJson
+      if pkg.alias.len == 0 and options.queryVersions:
+        info["versions"] = packageVersionsJson(pkg)
+      jsonOutput.add info
+    else:
+      echoPackage(pkg)
+      if pkg.alias.len == 0 and options.queryVersions:
+        echoPackageVersions(pkg)
+      echo(" ")
     found = true
     break forPkg
 
@@ -603,16 +611,33 @@ proc search(options: Options) =
 
   if not found:
     display("Error", "No package found.", Error, HighPriority)
+  elif options.jsonOutput:
+    echo jsonOutput
+
+proc echoList(pkgList: seq[Package], queryVersions: bool) =
+  for pkg in pkgList:
+    echoPackage(pkg)
+    if pkg.alias.len == 0 and queryVersions:
+      echoPackageVersions(pkg)
+    echo(" ")
+
+proc jsonList(pkgList: seq[Package], queryVersions: bool) =
+  let output = newJArray()
+  for i, pkg in pkgList:
+    let info = pkg.toJson
+    if queryVersions:
+      info["versions"] = packageVersionsJson(pkg)
+    output.add info
+  echo output
 
 proc list(options: Options) =
   if needsRefresh(options):
     raise newException(NimbleError, "Please run nimble refresh.")
   let pkgList = getPackageList(options)
-  for pkg in pkgList:
-    echoPackage(pkg)
-    if pkg.alias.len == 0 and options.queryVersions:
-      echoPackageVersions(pkg)
-    echo(" ")
+  if options.jsonOutput:
+    jsonList(pkgList, options.queryVersions)
+  else:
+    echoList(pkgList, options.queryVersions)
 
 proc listInstalled(options: Options) =
   var h = initOrderedTable[string, seq[string]]()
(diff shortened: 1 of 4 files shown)


src/nimblepkg/packageinfo.nim
```diff
@@ -425,6 +425,17 @@ proc echoPackage*(pkg: Package) =
     if pkg.web.len > 0:
       echo("  website:     " & pkg.web)
 
+func json*(pkg: Package): JsonNode =
```
dom96:
```suggestion
func toJson*(pkg: Package): JsonNode =
```

src/nimblepkg/packageinfo.nim
```diff
@@ -425,6 +425,17 @@ proc echoPackage*(pkg: Package) =
...
+  result = newJObject()
+  result["name"] = %pkg.name
+  result["url"] = %pkg.url
+  result["method"] = %pkg.downloadMethod
+  result["tags"] = %pkg.tags
+  result["description"] = %pkg.description
+  result["license"] = %pkg.license
```
dom96: This can be simplified to:

```nim
result = %*{
  "name": pkg.name,
  ...
}
```

src/nimblepkg/options.nim
```diff
@@ -106,6 +107,8 @@ Options:
   -v, --version                   Print version information.
   -y, --accept                    Accept all interactive prompts.
   -n, --reject                    Reject all interactive prompts.
+      --json                      Produce JSON formatted output when
```
dom96: This should be under the ``search`` and ``list`` commands above.

src/nimble.nim
```diff
@@ -547,12 +547,22 @@ proc search(options: Options) =
   if needsRefresh(options):
     raise newException(NimbleError, "Please run nimble refresh.")
   let pkgList = getPackageList(options)
-  var found = false
+  var
+    found = false
+    jsonOutput: JsonNode
+  if options.jsonOutput:
+    jsonOutput = newJArray()
```
dom96: This ``if`` is just noise, no problem in always initializing the JArray.

src/nimble.nim
```diff
@@ -547,12 +547,22 @@ proc search(options: Options) =
...
   template onFound {.dirty.} =
-    echoPackage(pkg)
-    if pkg.alias.len == 0 and options.queryVersions:
-      echoPackageVersions(pkg)
-    echo(" ")
+    if options.jsonOutput:
+      let info = pkg.json
+      if pkg.alias.len == 0 and options.queryVersions:
+        info["versions"] = packageVersionsJson(pkg)
```
dom96: Why not do this in the above `json` function?

src/nimble.nim
```diff
@@ -569,17 +579,35 @@ proc search(options: Options) =
...
 
+proc jsonList(pkgList: seq[Package], queryVersions: bool) =
+  echo "["
+  for i, pkg in pkgList:
+    let info = pkg.json
+    if queryVersions:
+      info["versions"] = packageVersionsJson(pkg)
+    echo info
+    if i != pkgList.high: echo ","
+  echo "]"
```
dom96: Please create a JArray and echo it.

src/nimblepkg/download.nim
```diff
@@ -286,6 +287,20 @@ proc echoPackageVersions*(pkg: Package) =
...
+  try:
+    case downMethod
+    of DownloadMethod.git:
+        let versions = getTagsListRemote(pkg.url, downMethod).getVersionList()
+        for v in values(versions):
+          result.add %v
+    of DownloadMethod.hg:
+      discard
+  except:
+    discard
```
dom96: You shouldn't swallow exceptions, please catch this on the ``getTagsListRemote`` and add an object with a "exception" key or something similar to the JsonNode.

src/nimblepkg/download.nim
```diff
@@ -286,6 +287,20 @@ proc echoPackageVersions*(pkg: Package) =
     echo("  versions:    (Remote tag retrieval not supported by " &
         pkg.downloadMethod & ")")
 
+proc packageVersionsJson*(pkg: Package): JsonNode =
+  result = newJArray()
+  let downMethod = pkg.downloadMethod.getDownloadMethod()
+  try:
+    case downMethod
+    of DownloadMethod.git:
+        let versions = getTagsListRemote(pkg.url, downMethod).getVersionList()
```
dom96: 2-space indent.

src/nimble.nim
ehmry: There is a problem with order of includes, and I prefer declaring `json` as a `func`.

src/nimble.nim
ehmry: The tradeoff is clean code or streaming output, but streaming output is not a requirement.

dom96: I'd prefer cleaner code.

src/nimble.nim
dom96: Order of includes? Not sure what you mean.

ehmry: There would be an import loop unless I move `toJson` to `download.nim`.

genotrance:
Ping

ehmry:
I now do my package parsing in an alternate frontend https://github.com/nix-community/flake-nimble/tree/master/src.

PR #709, opened 2019-09-21
Fixes #606

### LINKED ISSUE #606: before/after hooks for 'build' are not run
The following in a '.nimble' file:
```
before build:
   echo "before build"
after build:
   echo "after build"
```
should print:
```
before build
...
after build
```
when called with `nimble build` but they are never run.

zevv:
Is this enough to fix #606?

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -213,7 +213,7 @@ proc processDeps(pkginfo: PackageInfo, options: Options): seq[PackageInfo] =
   for i in reverseDeps:
     addRevDep(options.nimbleData, i, pkginfo)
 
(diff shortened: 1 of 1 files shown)


dom96:
Afraid not. ``buildFromDir`` is executed from various functions that aren't just called on `nimble build`. We need to do this in the ``build`` proc.

dom96:
Also these changes seem to have caused an infinite recursion in one of the tests.

dom96:
Thank you for giving it a shot :)

dom96:
Fixed this a bit differently.

PR #714, opened 2019-09-23
Enable Windows CI

genotrance:


.travis.yml
```diff
@@ -8,22 +9,43 @@ env:
...
   # This is the latest working Nim version against which Nimble is being tested
+  # Update in cache:directories as well
   - BRANCH=#2565d3d102efd21ba02ed1f3b96d892fe2637d2b
 
 cache:
   directories:
+    - "$HOME/.nimble/bin"
     - "$HOME/.choosenim/toolchains/nim-0.19.6"
     - "$HOME/.choosenim/toolchains/nim-0.20.2"
+    - "$HOME/.choosenim/toolchains/nim-#2565d3d102efd21ba02ed1f3b96d892fe2637d2b"
```
dom96: Wouldn't a simple:

```
  - "$HOME/.choosenim/toolchains/nim-$BRANCH"
```

Work here?

.travis.yml
```diff
@@ -8,22 +9,43 @@ env:
...
+        # Updating nimble to avoid failure with tests for #280 and #524
+        cd `./choosenim.exe show | grep Path | cut -f 8 -d " "` && ./koch.exe --latest nimble && cd -
+      else
+        export CHOOSENIM_CHOOSE_VERSION=$BRANCH
+        curl https://nim-lang.org/choosenim/init.sh -sSf > init.sh
+        sh init.sh -y
+      fi
+    else
+      choosenim $BRANCH -y
+    fi
```
dom96: Why move this to `before_script`?

.travis.yml
```diff
@@ -8,22 +9,43 @@ env:
...
+        #
+        # Using custom build with these PRs for Windows
+        curl -L -s "https://bintray.com/genotrance/binaries/download_file?file_path=choosenim.exe" -o choosenim.exe
+        curl -L -s "https://bintray.com/genotrance/binaries/download_file?file_path=libeay32.dll" -o libeay32.dll
+        curl -L -s "https://bintray.com/genotrance/binaries/download_file?file_path=ssleay32.dll" -o ssleay32.dll
+        ./choosenim.exe $BRANCH -y
+        cp ./choosenim.exe $HOME/.nimble/bin/.
+        # 0.19.6 and 0.20.2 have nimble with issue #665
+        # Updating nimble to avoid failure with tests for #280 and #524
+        cd `./choosenim.exe show | grep Path | cut -f 8 -d " "` && ./koch.exe --latest nimble && cd -
```
dom96: Nice, very thorough. But it seems very wrong to me that our test suite depends on another separate Nimble that we have in the PATH. This CI tests Nimble after all... can we do something about this instead of having these workarounds?

tests/tester.nim
```diff
@@ -922,6 +922,9 @@ test "Passing command line arguments to a task (#633)":
     check output.contains("Got it")
 
 suite "nimble run":
+  let
+    dotExeExt = if ExeExt.len != 0: "." & ExeExt else: ""
```
dom96: Can you use `changeFileExt` instead?

.travis.yml
genotrance: If only :( vars defined in the yaml don't work there.

.travis.yml
genotrance: This is all part of install and caching. Need ~/.nimble/bin in the path to check for choosenim. Else will have to download choosenim every run.

tests/tester.nim
genotrance: Updated

.travis.yml
genotrance: I've created https://github.com/nimble-test/issue280and524/pull/3 to fix this. Once you approve that, I can kick off the testing for this PR.

.travis.yml
dom96: huh, it's beneath ``install`` not ``before_script``. I guess I misread that.

.travis.yml
genotrance: PTAL - had to create https://github.com/nimble-test/issue280and524/pull/4 to do this right.

genotrance:
Please restart this CI once you merge https://github.com/nimble-test/issue280and524/pull/4.

src/nimble.nim
```diff
@@ -1063,7 +1063,14 @@ proc test(options: Options) =
...
       if canRemove:
-        removeFile(binFileName)
+        var retry = 2
+        while retry > 0:
+          try:
+            removeFile(binFileName)
+            retry = 0
+          except OSError:
+            retry -= 1
+            sleep(1)
```
dom96: Why do this?

PR #718, opened 2019-09-26
Allow git commit hashes to be specified in the nimble file

### LINKED ISSUE #715: requires syntax throws hg error
If you try to specify a git repo with a required version, nimble will think its pulling a mercurial repository.

Example:

```
requires "https://github.com/danielecook/BitVector#b8cc21271c90cca96ed31f5d5383711dc96a8d3f >= 0.4.11"
```

Results in this error:

```
     Error: 'hg' not in PATH.
```

It would be good to improve this error message or alternatively allow for version checking with specific commits OR disallow it and alert the user.

3n-k1:
This PR allows git commit hashes to be specified in the numble file, allowing nimble to treat git URLs like most other package managers do. Fixes #715.

I wasn't sure if there were any standards for the tester files, please let me know if you have any suggestions.


3n-k1:
Strange, that test passes on my machine. I'll try and find the issue and fix it.

3n-k1:
CI broke but it wasn't my fault
`Error: unhandled exception: Unable to query remote tags for https://github.com/vegansk/nimboost. Git returned: fatal: unable to access 'https://github.com/vegansk/nimboost/': Could not resolve host: github.com`

3n-k1:
Ah, yeah, that definitely makes more sense. If you'd like, I can just revert these changes and implement a quick check to make sure that no commit hash has been specified.

dom96:
Sure, might be better to just start with a fresh PR for that.

3n-k1:
Alright, will do!

3n-k1:
I do have a question though. What if someone wants to use the `nimble install` command to install from a specific hash instead of from the master branch, should that be supported, or no?

dom96:
it is supported. You just do ``nimble install pkg@#hash``

3n-k1:
That works fine if the package is registered with nimble, but what if a user needs to install a package that isn't in the nimble repository?

Araq:
Huh? How could nimble install a package that it doesn't know about?

dom96:
``nimble install <url_here>@#hash``

3n-k1:
Ah, thanks!

PR #719, opened 2019-09-26
Enable compile-time echos

3n-k1:
When using nimble to compile files, `echo`s in static blocks will be shown on stdout.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -251,9 +251,9 @@ proc buildFromDir(
       createDir(outputDir)
 
     try:
-      doCmd("\"" & getNimBin() & "\" $# --noNimblePath $# $# $# \"$#\"" %
+      doCmd("\"" & getNimBin() & "\" --hints=off $# --noNimblePath $# $# $# \"$#\"" %
             [pkgInfo.backend, nimblePkgVersion, join(args, " "), outputOpt,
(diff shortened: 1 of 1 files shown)


dom96:
And everything else too, the fact that Nimble doesn't show this output for those commands is by-design. If you'd like you can enable the output for `--verbose`.

In your PRs, please show me how you've tested this and the output that was returned for your test commands.

3n-k1:
That's why I added `--hints=off` to the `doCmd` on line 254. Adding the `--verbose` switch doesn't show the compile-time `echo`s.

I have a basic file s.nim that looks like this:
```nim
static:
  echo "Compiling 1. set of macros..."
```
and a basic s.nimble file that looks like this:
```nim
version       = "0.11.0"
author        = "Taylor hoff"
description   = "Test."
license       = "BSD"
bin = @["s"]
```
When running with `nimble run s`, the only output is:
```
  Verifying dependencies for s@0.11.0
   Building s/s using c backend
Compiling 1. set of macros...
```
I think allowing nimble to have compile-time `echo`s would be a good idea, but I understand if that's something you don't want.

dom96:
yeah, I'd rather not pick and choose like this. I think adding ``--hints=off`` might break ``nimble c `` as well.

3n-k1:
Alright :)

PR #728, opened 2019-10-20
useful git commit messages during publish

disruptek:
This replaces package publish commit messages of the form
```
  Added package "rando23"
```

With
```
  Added package "rando23"

  Tags: random rng
   Url: https://github.com/ralph/rando23
 Description:
  It makes you any random number you want from 0-23 inclusive, provided you wait long enough.
```
Which makes life much nicer for people who want to know what was added without having to jump through hoops to learn about the package.

### DIFF
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -94,11 +94,16 @@ proc createFork(a: Auth) =
     raise newException(NimbleError, "Unable to create fork. Access token" &
                        " might not have enough permissions.")
 
-proc createPullRequest(a: Auth, packageName, branch: string): string =
+proc createPullRequest(a: Auth, packageName, branch: string, text = ""): string =
   display("Info", "Creating PR", priority = HighPriority)
+  let js = %* {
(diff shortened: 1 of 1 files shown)


dom96:
You can see this information in the diff though, why include it in the message?

disruptek:
The diff isn't included in the email, so there's already a barrier of entry.  It may result in my wasting a visit to the web just to find out that I'm not interested.  To me, that's even worse.

If I want to see the patch, I can open the link from my email and see it in a raw format, wherein I cannot click on the link to visit the GitHub page.  At least this way I can read tags and description.

If I visit the GitHub page for the pull request, I think have to click on the link for the commit before I see any useful description for the package.  I _still_ don't have a link to the url, though, so now *I have to copy and paste it* to the location bar using a selection from my pointer.  I can't even triple-click the url to select it because my IFS will either pick up the entire JSON line or select only one word from the url.

Is this really the set of hoops you want me to jump through just to LEARN about new Nim software?

If there's a better way, I'm all ears.

PR #729, opened 2019-10-22
add vcs to the new created project

yuchunzhou:


nimble.nimble
```diff
@@ -11,7 +11,7 @@ installExt = @["nim"]
 
 # Dependencies
 
-requires "nim >= 0.13.0"
+requires "nim >= 0.13.0", "gitapi"
```
SolitudeSF: nimble shouldn't depend on external packages.

src/nimble.nim
```diff
@@ -858,6 +860,19 @@ js   - Compile using JavaScript backend.""",
     pkgRoot
   )
 
+  # Add git vcs to the new project
+  let repo = createGitRepo(path=pkgRoot)
+  discard repo.gitCommand("init")
+
+  var fd = open(joinPath(pkgRoot, ".gitignore"), fmWrite)
```
SolitudeSF: library packages dont produce executables except for tests. and nothing ever produces `*.out` unless you explicitly do that.

src/nimble.nim
```diff
@@ -858,6 +860,19 @@ js   - Compile using JavaScript backend.""",
...
+  fd.write(pkgName & "\n" & "*.out\n")
+  fd.close()
+
+  # Add a default xxx.nim.cfg file
+  fd = open(joinPath(pkgRoot, pkgSrcDir, pkgName & ".nim.cfg"), fmWrite)
+  fd.write("--d:release\n")
```
SolitudeSF: you cant apply `release/danger` in `cfg` and it absolutely shouldn't be the default for no reason.

nimble.nimble
yuchunzhou: ok, if the git init function is accepted, i can refactor the code later.

src/nimble.nim
yuchunzhou: @SolitudeSF why not specify a release version in the .nim.cfg file, i found that there are many nim users do not know how to input the compiler flags when compile a project, i thing it's a good beginning to create a default .nim.cfg when creating a new project.

SolitudeSF: if they dont know about compiler flags, they for sure dont know about configuration files. and even if it was a good idea, it doesnt even work. you cant apply release flag from user configuration.

dom96:
This is a cool idea, although I'm not sure we should be implicitly initializing a git repo when creating a new Nimble project. Some might not use git but mercurial instead.

How about you put these behind a ``--git`` flag?

yuchunzhou:
@dom96 ok, i will try to update it.

src/nimble.nim
yuchunzhou: ok, i revert it.

Araq: @SolitudeSF It's high time we fix that.

yuchunzhou:
hi @dom96, i have added a `--vcs:git` option to init flag, the pull request supports git and hg both now, please review it.

src/nimble.nim
yuchunzhou: @SolitudeSF i think, if nim needs some compile flags, we should write them into a .cfg file, maybe it's not a good choice to write the compiler flags in the comand line. If all of us agree with the .cfg feature, i can add them before finish initializing a new project.

src/nimblepkg/options.nim
```diff
@@ -80,6 +81,8 @@ Commands:
   init         [pkgname]          Initializes a new Nimble project in the
                                   current directory or if a name is provided a
                                   new directory of the same name.
+               [--vcs:git]        Add git or hg vcs to the new nimble project,
```
dom96: I would prefer explicit `--git` and `--hg` flags. Also ``git`` shouldn't be default, neither should be default.

src/nimble.nim
```diff
@@ -858,6 +860,22 @@ js   - Compile using JavaScript backend.""",
     pkgRoot
   )
 
+  # Add vcs to the new nimble project
+  let vcsBin = options.action.vcsOption
+  if findExe(vcsBin, true).len != 0:
```
dom96: This should be checked at the beginning of the function, if it doesn't exist then you should throw a NimbleError.

The actual logic you have here should still happen at the end though.

yuchunzhou:
@dom96  ok, i updated the request's code and added `--git` and `--hg` flags for `nimble init` command.

src/nimble.nim
yuchunzhou: ok, i  put the check code to the beginning of the init function and the actual logic is still at the end of the init function

src/nimblepkg/options.nim
yuchunzhou: yeah, i agree with you, i add both `--git` and `--hg` flags and if user inputs more than one flags, the later will overwrite the before ones.

yuchunzhou:
@dom96 I find Nim is friendly to Windows platform, how about add a `--svn` flag to `nimble init` command?

PR #736, opened 2019-11-03
Add '==' as operator for pinning versions.

itmuckel:


itmuckel:
I'm not quite sure about the structure of the tests. It might get messy with separate folders for gitlab issues. 🤔 
The test I added should make sure that both equality operators behave the same. Feel free to adapt it, if you find something missing here.

itmuckel:
Okay, I have no idea why the test "can list" fails now. Trying it locally on my machine works.

dom96:
> Okay, I have no idea why the test "can list" fails now. Trying it locally on my machine works.

That's because: https://github.com/nim-lang/packages/commit/54b2556c518519d7ef1e7b81406bbe045668cfbc

PR #744, opened 2019-11-19
Capitalize the "D" in srcDir to avoid errors when using styleChecks

### LINKED ISSUE #740: default nimble file generation breaks when used with --styleChecks:error
It turns out that `srcDir` breaks the case sensitivity check from `styleChecks`, and should either be generated as `srcdir`, or the original declaration should be changed to camel case.

Steps to reproduce:
```bash
nimble init mytestproj
cd mytestproj
echo '--styleChecks:error' > config.nims
nimble build
```

3n-k1:
Fixes #740

When using a config.nims containing `--styleCheck:error`, `srcDir` will no longer throw an error. Originally, nimscriptapi.nim defined the variable as `srcdir`, without the capitalized D. As nimble generates a nimble file containing `srcDir` (with the capital D), the compiler would error out.

I know this is a small change, but since everywhere else refers to the variable as `srcDir`, I think it's a good change.

### DIFF
--- a/src/nimblepkg/nimscriptapi.nim
+++ b/src/nimblepkg/nimscriptapi.nim
@@ -17,7 +17,7 @@ var
   author*: string      ## The package's author.
   description*: string ## The package's description.
   license*: string     ## The package's license.
-  srcdir*: string      ## The package's source directory.
+  srcDir*: string      ## The package's source directory.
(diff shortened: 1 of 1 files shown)


3n-k1:
Awesome, thanks for the merge!

PR #761, opened 2020-01-09
Run can work without additional option if only one bin in .nimble

inv2004:
if you have only one bin in .nimble.

before:
  you have to specific executable every time: nimble run my_tool

after PR:
  you can just: nimble run

additional fix:
before:
  windows parsed empty option not correct => ".exe"

after PR:
  it will raise correct exception "Please specify a binary to run" if binary not defined in opts or config

### DIFF
--- a/src/nimblepkg/options.nim
+++ b/src/nimblepkg/options.nim
@@ -89,8 +89,8 @@ Commands:
   uninstall    [pkgname, ...]     Uninstalls a list of packages.
                [-i, --inclDeps]   Uninstall package and dependent package(s).
   build        [opts, ...] [bin]  Builds a package.
-  run          [opts, ...] bin    Builds and runs a package.
-                                  A binary name needs
+  run          [opts, ...] [bin]  Builds and runs a package.
+                                  If several binaries defined then binary name needs
                                   to be specified after any compilation options,
                                   any flags after the binary name are passed to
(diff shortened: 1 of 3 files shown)


inv2004:
@dom96 
Thank you for pointing me to the tests. Which helps me to find another problem of passing parameters to the "default" binary.

I added "--" parameter to replace default executable _if it is necessary_
for example:
nimble run --debug // default binary works fine
nimble run --debug check // it thinks that we set "check" to binary, but it is not correct
nimble run -- --debug check // fix previous problem

src/nimblepkg/options.nim
```diff
@@ -89,8 +89,8 @@ Commands:
   uninstall    [pkgname, ...]     Uninstalls a list of packages.
                [-i, --inclDeps]   Uninstall package and dependent package(s).
   build        [opts, ...] [bin]  Builds a package.
-  run          [opts, ...] bin    Builds and runs a package.
-                                  A binary name needs
+  run          [opts, ...] [bin]  Builds and runs a package.
+                                  If several binaries defined then binary name needs
```
Yardanico: I think this english is a little bit wrong, IMO "If there are several binaries defined, then the binary name " would be a bit better

inv2004: updated, Was trying to not make it too long

PR #768, opened 2020-02-05
nimscript{wrapper,api}: don't copy generated script to package directory

alaviss:
This allows nimble to process read-only package directories.

Pretty much a draft for now since I don't have any way to test for this implementation shortcomings.

In short, instead of copying the generated script to the package dir, I run it directly in the nimble cache with paths pointing to the package dir as well as some magic to fool the script into thinking that it's running inside the package dir.

/cc @genotrance

### DIFF
--- a/src/nimblepkg/nimscriptwrapper.nim
+++ b/src/nimblepkg/nimscriptwrapper.nim
@@ -37,30 +37,19 @@ proc writeExecutionOutput(data: string) =
   display("Info", data)
 
 proc execNimscript(
-  nimsFile, projectDir, actionName: string, options: Options, isHook: bool
+  nimsFile, projectDir, actionName: string, options: Options, isHook: bool,
+  nimbleFile = ""
 ): tuple[output: string, exitCode: int, stdout: string] =
   let
-    nimsFileCopied = projectDir / nimsFile.splitFile().name & "_" & getProcessId() & ".nims"
     outFile = getNimbleTempDir() & ".out"
 
-  let
-    isScriptResultCopied =
-      nimsFileCopied.fileExists() and
-      nimsFileCopied.getLastModificationTime() >= nimsFile.getLastModificationTime()
-
-  if not isScriptResultCopied:
-    nimsFile.copyFile(nimsFileCopied)
-
-  defer:
-    # Only if copied in this invocation, allows recursive calls of nimble
-    if not isScriptResultCopied and options.shouldRemoveTmp(nimsFileCopied):
-        nimsFileCopied.removeFile()
-
   var cmd = (
-    "nim e $# -p:$# $# $# $#" % [
+    "nim e $# -p:$# -p:$# $# $# $# $#" % [
       "--hints:off --verbosity:0",
       (getTempDir() / "nimblecache").quoteShell,
-      nimsFileCopied.quoteShell,
+      projectDir.quoteShell,
+      nimsFile.quoteShell,
+      if nimbleFile.len > 0: nimbleFile.quoteShell else: "",
       outFile.quoteShell,
       actionName
     ]
@@ -76,7 +65,6 @@ proc execNimscript(
         cmd &= ":" & val.quoteShell()
 
   displayDebug("Executing " & cmd)
-
   if needsLiveOutput(actionName, options, isHook):
     result.exitCode = execCmd(cmd)
   else:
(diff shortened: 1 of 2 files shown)


genotrance:
Using `--path` will solve the issue with relative imports but won't cover cases where the user does other relative file operations like cpFile, rmDir and the like.

Even simple things in the [nimterop.nimble](https://github.com/nimterop/nimterop/blob/master/nimterop.nimble#L27) file - compiling `nimterop/toast` or `tests/tnimterop_cpp.nim` - won't work.

alaviss:
@genotrance 

I actually tried what you said and... it works without any further patches. But I'm adding a patch anyway to be sure of the state for all weird cases.

genotrance:
LGTM - let's see what dom96 says.

Adding important_packages testing to nimble will be great.

src/nimblepkg/nimscriptwrapper.nim
```diff
@@ -77,12 +66,23 @@ proc execNimscript(
 
   displayDebug("Executing " & cmd)
 
+  var poptions = {poEvalCommand}
```
dom96:
```suggestion
  var processOpts = {poEvalCommand}
```

src/nimblepkg/nimscriptwrapper.nim
```diff
@@ -37,30 +37,19 @@ proc writeExecutionOutput(data: string) =
   display("Info", data)
 
 proc execNimscript(
-  nimsFile, projectDir, actionName: string, options: Options, isHook: bool
+  nimsFile, projectDir, actionName: string, options: Options, isHook: bool,
+  nimbleFile = ""
```
dom96: Why a default here? Doesn't every invokation of this have a ``nimbleFile`` to supply? and shouldn't it always be supplied?

alaviss: I originally thought that `execNimscript` was meant was a nimscript executing wrapper for any nimscript file. Turns out that it was just meant to execute nimble scripts. I don't know why this don't just take a nimble file instead, every invocation I found is in the form of `getNimsFile` then `execNimscript`.

genotrance: Nim doesn't like `.nimble` files, it has to be `.nims`. Further, we add some stuff to the `.nims` file generated to retain backward compatibility.

https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/nimscriptwrapper.nim#L117

alaviss: I know, what I mean is that this step can be done in `execNimscript` instead of a separated `getNimsFile` proc because `execNimscript` can only ever execute scripts created by `getNimsFile`.

genotrance: `getIniFile()` also uses it, was one way to reuse.

dom96: So, can we resolve this?

dom96:
@genotrance are you happy with this being merged?

genotrance:
This looks good to be, I believe @alaviss was going to test all important packages to ensure it didn't break anything.

alaviss:
The biggest issue right now is that the project configuration (ie. `nim.cfg`, `config.nims`) will not be respected by nimble while executing the `.nimble` script after this patch. And currently I don't see any possible solutions for this scenario...

genotrance:
> The biggest issue right now is that the project configuration (ie. `nim.cfg`, `config.nims`) will not be respected by nimble while executing the `.nimble` script after this patch. And currently I don't see any possible solutions for this scenario...

So nimterop recently added functionality to get the nim configuration for a directory since it calls nim itself to compile its plugins. This code can be reused to give nimble the same ability.

https://github.com/nimterop/nimterop/blob/master/nimterop/nimconf.nim

The code is capable of detecting the configuration and writing it to a cfg file so it can be used to detect the config of the .nimble directory and writing it next to the nims file in temp.

It only covers `--path`, `--nimblePath` and `--nimcache` for now. If more flags are needed, we can consider expanding it. It uses `nim dump` so any other flags might need compiler changes.

If this is worthwhile, I can make it a standalone package.

alaviss:
Does `dump.format=json` exists in older Nim?

genotrance:
It provides limited info on 0.20.2 and 1.0.6. The code tries to handle it as much as possible.

alaviss:
We will have to settle with whether we want full Nim configuration support for running nimble script.

genotrance:
Closing since #826 covers these changes.

PR #770, opened 2020-02-05
Use travis gist

genotrance:


genotrance:
@dom96 - this is ready to go.

.travis.yml
```diff
@@ -8,37 +8,18 @@ language: c
...
-        curl -L -s "https://bintray.com/genotrance/binaries/download_file?file_path=ssleay32.dll" -o ssleay32.dll
-        ./choosenim.exe $BRANCH -y
-        cp ./choosenim.exe $HOME/.nimble/bin/.
-      else
-        export CHOOSENIM_CHOOSE_VERSION=$BRANCH
-        curl https://nim-lang.org/choosenim/init.sh -sSf > init.sh
-        sh init.sh -y
-      fi
-    fi
+  - curl https://gist.github.com/genotrance/fb53504a4fba88bc5201d3783df5c522/raw/travis.sh -LsSf -o travis.sh
```
dom96: I'm curious why you're downloading portablegit.exe for Windows. Shouldn't travis already have git preinstalled?

genotrance: Yes, it is to work around a cmake issue. Travis has a space in the path to git/sh.exe causing cmake to fail.

Nimble doesn't need cmake but the script is used across multiple CI including choosenim so doesn't hurt to have it.

PR #775, opened 2020-02-13
Fix #741 - colored Nim output

### LINKED ISSUE #741: Nimble build errors aren't as friendly as nim
From what I understand, the standard practice is to build nim projects with the `nimble build` command. However, this strips colors from the compiler output and adds some extra indentation. Given that a large percent of the time `nimble build` errors will contain output from `nim`, I think it'd make sense to have it display errors nicely. Here's a comparison of with `nimble build` vs with `nim c`:

Nimble:
image

Nim:
image

Let me know if this makes sense.

genotrance:


genotrance:
@dom96 - can you please review and merge?

PR #779, opened 2020-03-08
Fix #778, use platform specific newlines in init

### LINKED ISSUE #778: nimble init uses \n instead of platform specific linebreak
https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/init.nim

Multiline strings are used here which seem to always give `\n` no matter what the source file uses. This creates files with just \n and not \r\n on Windows.

Possible solution is `p` template mentioned [here](https://github.com/nim-lang/Nim/issues/7202).

hlaaftana:


genotrance:
Looks good. @dom96?

src/nimblepkg/init.nim
```diff
@@ -14,6 +14,8 @@ type
     pkgNimDep: string
     pkgType: string
 
+template p(s: string{lit}): string = static(s.replace("\n", "\p"))
```
dom96, line 17: Hold up, Nim doesn't use platform-specific newlines in multi-line string literals? That seems like a bug to me, and indeed the reason Nimble used \n here is because in the past it had the same meaning as `\p`.

Can you create an issue on the Nim repo and ask about this?

Araq: Nim always uses LF for ``\n`` and that is fine since every Windows tool in widespread use understands LF.

Araq:
It's better the way it is.

Issue #799, opened 2020-05-05
nimble build doesn't produce deterministic results


samdmarshall:
this is filed in conjunction with https://github.com/FedericoCeratto/nim-httpauth/issues/27, due to nimble not correctly installing a package based on directory format and file contents. the issue comes from nimble not installing all of the necessary files, so when importing the package into your own code, it will cause a build error due to not being able to locate specific imports. however, this only is useful in demonstrating that the current behavior of the `nimble build` command doesn't produce reproducible/deterministic results.

when a build (an invocation of `nimble build`) is performed, all of the required package dependencies are checked and, if necessary, installed locally. when installing a package, the package is cloned into a temp directory, then the necessary components are installed under the nimble directory (typically `~/.nimble/`). however, when performing a build, nimble uses the temp directory as the package path instead of the path to where it was installed under the nimble directory. 

i've included two build logs here, that you can see the package paths used when building the target differ depending if a dependency package is already installed or not.
 
<details>
<summary>Build log that requires package installation</summary>
<pre><code>
$ nimble --verbose build
Reading config file at /home/demi/.config/nimble/nimble.ini
  Verifying dependencies for kisuru@0.1.0
    Reading personal package list
    Reading official package list
   Checking for jester@>= 0.4.3
      Info: Dependency on jester@>= 0.4.3 already satisfied
  Verifying dependencies for jester@0.4.3
    Reading personal package list
    Reading official package list
   Checking for httpbeast@>= 0.2.2
      Info: Dependency on httpbeast@>= 0.2.2 already satisfied
  Verifying dependencies for httpbeast@0.2.2
    Reading personal package list
    Reading official package list
   Checking for https://github.com/timotheecour/asynctools@#pr_fix_compilation
      Info: Dependency on https://github.com/timotheecour/asynctools@#pr_fix_compilation already satisfied
  Verifying dependencies for asynctools@#pr_fix_compilation
    Reading personal package list
    Reading official package list
   Checking for https://github.com/timotheecour/asynctools@#pr_fix_compilation
      Info: Dependency on https://github.com/timotheecour/asynctools@#pr_fix_compilation already satisfied
  Verifying dependencies for asynctools@#pr_fix_compilation
    Reading personal package list
    Reading official package list
   Checking for httpauth@#head
 Installing httpauth@#head
    Reading personal package list
    Reading official package list
Downloading https://github.com/FedericoCeratto/nim-httpauth using git
[...]

PR #800, opened 2020-05-05
fix for issue #799

samdmarshall:
updates path to nimble package before assignment as result of package
installation, thus using the correct package path instead of the temporary path
the package was cloned into. (see issue #799 for more details)

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -456,11 +456,13 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
   # processDeps).
   saveNimbleData(options)
 
+  # update package path to point to installed directory rather than the temp directory
+  pkgInfo.myPath = dest
+
   # Return the dependencies of this package (mainly for paths).
   result.deps.add pkgInfo
   result.pkg = pkgInfo
   result.pkg.isInstalled = true
-  result.pkg.myPath = dest
 
   display("Success:", pkgInfo.name & " installed successfully.",
           Success, HighPriority)
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -458,11 +458,11 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
 
   # update package path to point to installed directory rather than the temp directory
   pkgInfo.myPath = dest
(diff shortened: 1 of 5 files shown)


src/nimble.nim
```diff
@@ -456,11 +456,13 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
   # processDeps).
   saveNimbleData(options)
 
+  # update package path to point to installed directory rather than the temp directory
+  pkgInfo.myPath = dest
+
   # Return the dependencies of this package (mainly for paths).
   result.deps.add pkgInfo
   result.pkg = pkgInfo
   result.pkg.isInstalled = true
```
samdmarshall, line 465: on closer inspection this line, `.isInstalled = true`, probably should also be done before the `pkgInfo` var is added to `result.deps` and set to `result.pkg`. though this isn't necessary for fixing the issue at hand, it is probably the correct thing to do. i can add another commit to this PR that makes this change too if you think that is appropriate, @dom96.

samdmarshall:
yeah, that should be possible. personally i think the test case gere should be that all packages that are passed to the compiler via the `--path` flag are sub-directories of the currently configured `--nimblepath`. does that sound suitable to you?

also: what do you think of my comment regarding the other property that gets updated? currently it is only "correct" in one of the two assignment locations, which is the source of this bug in the first place - that should probably be moved to be updated like this too.

also, also: question for my own curiousity: would it be acceptable to make additional future PRs that add (only) code comments to the existing codebase? i know i filed the issue and then this PR in fairly quick succession, but it took me a good chunck of time to track down the fix. part of that was due to unfamiliar code, but understanding what was being done also wasn't easy. i may not have time for major feature work but i could contribute documentation for the existing code to make PRs easier to make for everyone - but think it is only proper to get some form of permission to start doing that first.

samdmarshall:
@Araq / @dom96 could one of you retrigger the build to run, please? i don't believe my changes are the cause of the one error here that is holding up being able to merge this.

dom96:
Looks good, thanks! Hope you've verified that your test fails before your changes btw.

PR #802, opened 2020-05-19
Fixes #798 (relying on undefined behaviour)

### LINKED ISSUE #798: Nimble doesn't work with --gc:arc
Just to store the info:

`nimble.nim:416` - need to add nosinks like

```
proc (file: string) {.nosinks.} =
```
`download.nim:113` - need to do the same

```
.map(proc(s: string): tuple[ver: Version, tag: string] {.nosinks.} =
```

and the most important one - Nimble relies on the fact that after catching the exception the variable initialized in the "try" block still contains useful info (for validation), as shown in `packageparser.nim`, or for shorter example:
```nim
type
  PackageInfo* = object
    myPath*: string ## The path of this .nimble file

proc readPackageInfo(): PackageInfo = 
  result = PackageInfo(mypath: "testlol")
  if result.mypath == "testlol":
    raise newException(ValueError, "testlol")
  echo result.mypath

proc getPkgInfoFromFile*(): PackageInfo =
  ## Reads the specified .nimble file and returns its data as a PackageInfo
  ## object. Any validation errors are handled and echoed as warnings.
  try:
    result = readPackageInfo()
  except ValueError:
    echo "error"
  echo "getPkgInfoFromFile:"
  echo result.mypath

proc getPkgInfo*(): PackageInfo = 
  return getPkgInfoFromFile()

proc downloadPkg*() = 
  let pkginfo = getPkgInfo()
  echo pkginfo.mypath

downloadPkg()
```
I think this can be considered as relying on undefined behaviour, so this probably should be changed (store PackageInfo as the field in ValidationError ?)

Yardanico:
This PR fixes  #798. 
I'm not sure if this is the best way, but it requires the least amount of changes to the code. It's required for https://github.com/nim-lang/Nim/pull/14390 .

I'm not 100% sure that I fixed all places where that behaviour was used, so please comment if you find any issues.

### DIFF
--- a/src/nimblepkg/packageparser.nim
+++ b/src/nimblepkg/packageparser.nim
@@ -15,6 +15,7 @@ type
   ValidationError* = object of NimbleError
     warnInstalled*: bool # Determines whether to show a warning for installed pkgs
     warnAll*: bool
+    pkgInfo*: PackageInfo
 
 const reservedNames = [
   "CON",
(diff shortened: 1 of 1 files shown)


src/nimblepkg/packageparser.nim
```diff
@@ -464,6 +464,7 @@ proc getInstalledPkgs*(libsDir: string, options: Options):
           pkg = readPackageInfo(nimbleFile, options, onlyMinimalInfo=false)
         except ValidationError:
           let exc = (ref ValidationError)(getCurrentException())
+          pkg = exc.pkgInfo
```
Yardanico, line 467: By the way - this procedure (``getInstalledPkgs``) is not used anywhere, is that okay?

src/nimblepkg/packageparser.nim
```diff
@@ -41,16 +42,21 @@ const reservedNames = [
...
+
+proc raiseNewValidationError(info: PackageInfo, msg: string,
+                             warnInstalled: bool, hint = "", warnAll = false) =
+  raise info.newValidationError(msg, warnInstalled, hint, warnAll)
 
 proc raiseNewValidationError(msg: string, warnInstalled: bool,
-                             hint: string = "", warnAll = false) =
-  raise newValidationError(msg, warnInstalled, hint, warnAll)
+                             hint = "", warnAll = false) =
+  raise PackageInfo().newValidationError(msg, warnInstalled, hint, warnAll)
```
Yardanico, line 59: This might not look correct - but in fact it is - it's only used this way in ``validatePackageName`` which is only called from ``readPackageInfo`` where it's not covered in a try/except block so it won't make sense (and will require changing more code) to pass PackageInfo() there.
I also could've made a new newValidationError proc without info: PackageInfo, but I wanted to keep my changes as minimal as possible

PR #809, opened 2020-06-10
Add warning  message when cloning the HEAD.

ArMour85:
In answer of this [issue](https://github.com/nim-lang/packages/issues/1051), it is important to indicate that the developer use a package without git release tag.

This message, I hope, can also generate more serious in development process and improve the quality. In the common world, a release version has a git tag.

### DIFF
--- a/src/nimblepkg/download.nim
+++ b/src/nimblepkg/download.nim
@@ -205,7 +205,7 @@ proc doDownload(url: string, downloadDir: string, verRange: VersionRange,
                   onlyTip = not options.forceFullClone)
       else:
         # If no commits have been tagged on the repo we just clone HEAD.
-        display("Warning:", "No corresponding tag found with required version. We grab HEAD.", Warning, 
+        display("Warning:", "The package has no tagged releases, downloading HEAD instead.", Warning, 
                   priority = HighPriority)
         doClone(downMethod, url, downloadDir) # Grab HEAD.
     of DownloadMethod.hg:
--- a/src/nimblepkg/download.nim
+++ b/src/nimblepkg/download.nim
@@ -205,6 +205,8 @@ proc doDownload(url: string, downloadDir: string, verRange: VersionRange,
                   onlyTip = not options.forceFullClone)
       else:
         # If no commits have been tagged on the repo we just clone HEAD.
+        display("Warning:", "No corresponding tag found with required version. We grab HEAD.", Warning, 
+                  priority = HighPriority)
         doClone(downMethod, url, downloadDir) # Grab HEAD.
     of DownloadMethod.hg:
       doClone(downMethod, url, downloadDir, onlyTip = not options.forceFullClone)
@@ -216,6 +218,9 @@ proc doDownload(url: string, downloadDir: string, verRange: VersionRange,
           display("Switching", "to latest tagged version: " & latest.tag,
                   priority = MediumPriority)
           doCheckout(downMethod, downloadDir, latest.tag)
+      else:
+        display("Warning:", "No corresponding tag found with required version.", Warning, 
+                  priority = HighPriority)
 
 proc downloadPkg*(url: string, verRange: VersionRange,
                  downMethod: DownloadMethod,
--- a/src/nimblepkg/download.nim
+++ b/src/nimblepkg/download.nim
@@ -219,7 +219,7 @@ proc doDownload(url: string, downloadDir: string, verRange: VersionRange,
                   priority = MediumPriority)
           doCheckout(downMethod, downloadDir, latest.tag)
       else:
-        display("Warning:", "No corresponding tag found with required version.", Warning, 
+        display("Warning:", "The package has no tagged releases, downloading HEAD instead.", Warning, 
                   priority = HighPriority)
 
 proc downloadPkg*(url: string, verRange: VersionRange,


ArMour85:
The warning message is displayed only if the repo hasn't tags when a version is needed. I made some tests:

**zip** package hasn't tags:
```
./src/nimble install zip@#head
Downloading https://github.com/nim-lang/zip using git
  Verifying dependencies for zip@#head
 Installing zip@#head
   Success: zip installed successfully.

./src/nimble install zip
Downloading https://github.com/nim-lang/zip using git
   Warning: No corresponding tag found with required version. We grab HEAD.
  Verifying dependencies for zip@0.2.1
 Installing zip@0.2.1
    Prompt: zip@0.2.1 already exists. Overwrite? [y/N]
    Answer: y
   Success: zip installed successfully.
```

**karax** package has tags:
```
nimble install karax@#head
Downloading https://github.com/pragmagic/karax using git
   Warning: Package 'karax' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'karax', but file 'button.nim' is in a directory named 'examples' instead. This will be an error in the future.
      Hint: If 'examples' contains source files for building 'karax', rename it to 'karax'. Otherwise, prevent its installation by adding `skipDirs = @["examples"]` to the .nimble file.
  Verifying dependencies for karax@#head
 Installing karax@#head
   Building karax/karax/tools/karun using c backend
   Success: karax installed successfully.

nimble install karax
Downloading https://github.com/pragmagic/karax using git
   Warning: Package 'karax' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'karax', but file 'button.nim' is in a directory named 'examples' instead. This will be an error in the future.
      Hint: If 'examples' contains source files for building 'karax', rename it to 'karax'. Otherwise, prevent its installation by adding `skipDirs = @["examples"]` to the .nimble file.
  Verifying dependencies for karax@1.1.2
 Installing karax@1.1.2
   Building karax/karax/tools/karun using c backend
   Success: karax installed successfully.
```

ArMour85:
I can maybe improve the message.

genotrance:
I'll suggest that message should always be shown even if the user explicitly asks for #head. It should further explain the real problem with #head - if the user installs `pkg1@#head` at some point and then later on `pkg2` requires `pkg1@v0.5.0` which is newer than when `pkg1@#head` was cloned, it won't ever get downloaded by Nimble. `pkg2` will try to build with the old `pkg1@#head` and the build might fail leaving the user guessing what the problem is.

Nimble will assume that `pkg1@#head` in the repo is the latest so it totally throws off dependency version checking. Even if you force install `pkg1@v0.5.0`, I think `#head` will be treated as newer.

Maybe we should always grumble that a package version `#head` is in use and warn the user that it could be out of date. Then of course, the user will want a way to update that to the latest `#head` and the headache never ends.

ArMour85:
@genotrance I totally agree with you to add a warning messages when a user use `#head`. I can add it with the warning I have already added.

dom96:
> Nimble will assume that pkg1@#head in the repo is the latest so it totally throws off dependency version checking. Even if you force install pkg1@v0.5.0, I think #head will be treated as newer.

That's not quite the case I don't think. The distinction here is with how Nim and Nimble handle dependencies. The Nim compiler does not read .nimble files to check versions but Nimble should and does AFAIK.

ArMourMag:
> > Nimble will assume that pkg1@#head in the repo is the latest so it totally throws off dependency version checking. Even if you force install pkg1@v0.5.0, I think #head will be treated as newer.
> 
> That's not quite the case I don't think. The distinction here is with how Nim and Nimble handle dependencies. The Nim compiler does not read .nimble files to check versions but Nimble should and does AFAIK.

Sorry, but I don't understand your answer about the relationship between Nim and Nimble to manage dependencies (I don't very smart). Yes Nim doesn't care about .nimble files but we talk about Nimble when it uses to install, build or run a project. Personally, I rarely run Nim to build my project but rather `nimble build`. In this cases, using `#head` is a problem as @genotrance says.

ArMour85:
> > Nimble will assume that pkg1@#head in the repo is the latest so it totally throws off dependency version checking. Even if you force install pkg1@v0.5.0, I think #head will be treated as newer.
> 
> That's not quite the case I don't think. The distinction here is with how Nim and Nimble handle dependencies. The Nim compiler does not read .nimble files to check versions but Nimble should and does AFAIK.

Sorry, but I don't understand your answer about the relationship between Nim and Nimble to manage dependencies (I don't very smart). Yes Nim doesn't care about .nimble files but we talk about Nimble when it uses to install, build or run a project. Personally, I rarely run Nim to build my project but rather `nimble build`. In this cases, using `#head` is a problem as @genotrance says.

src/nimblepkg/download.nim
```diff
@@ -216,6 +218,9 @@ proc doDownload(url: string, downloadDir: string, verRange: VersionRange,
           display("Switching", "to latest tagged version: " & latest.tag,
                   priority = MediumPriority)
           doCheckout(downMethod, downloadDir, latest.tag)
+      else:
+        display("Warning:", "No corresponding tag found with required version.", Warning, 
+                  priority = HighPriority)
```
dom96, line 223: Make the message here the same as above.

src/nimblepkg/download.nim
```diff
@@ -205,6 +205,8 @@ proc doDownload(url: string, downloadDir: string, verRange: VersionRange,
                   onlyTip = not options.forceFullClone)
       else:
         # If no commits have been tagged on the repo we just clone HEAD.
+        display("Warning:", "No corresponding tag found with required version. We grab HEAD.", Warning,
```
dom96, line 208:
```suggestion
        display("Warning:", "The package has no tagged releases, downloading HEAD instead.", Warning, 
```

dom96 (approved):
See remarks, but seems fine otherwise.

ArMour85:
Bump.

Can you merge it please ?

dom96:
Can you make my suggested changes? :)

ArMour85:
Sorry, it's good now.

src/nimblepkg/download.nim
dom96: You forgot this one :)

PR #825, opened 2020-07-24
Caching with includes

genotrance:


dom96:
Curious about the rationale for this. Can you explain?

genotrance:
I removed nimscriptapi caching which obviously breaks when nimble is installed away from its source code, so reverting that back. However, I'm making it unique to the nimble binary so that using older and newer versions of nimble don't break because of an incompatible nimscriptapi in /tmp.

In addition, I am doing `include <path/to/nimble/file>` so we will get better errors with line numbers in the nimble file. In the process, we no longer need to check if the nimble file changes since it is simply being included. Still need to make that specific code change.

Next step is to continue experimenting with what @alaviss had in #768. After discussing with him on IRC, seems like we shouldn't really care about `nim.cfg` and `config.nims` since Nimble pre `nim e` probably didn't care about those files anyway.

What remains is verifying that commands like `gorgeEx()` still work once we stop copying files over to the project directory.

dom96:
Sounds good :)

PR #828, opened 2020-07-27
Fix #382, fix #749 - improve Nim output

### LINKED ISSUE #382: Do not swallow compiler output
I'd like to see the progress going when my project is being built, and I also like to see the compilation stats (memory, time) Nim outputs when it finishes. But nimble swallows all of that and just silently exits when build process is complete.
It was fine before the fancy coloring was added to nimble output. So for some reason a useful feature was replaced with cosmetics at that point.

### LINKED ISSUE #749: Hide hints during tests running
Hello! Is there any way to hide hints and other messages which not relates to test suites and cases?

Running just with `nimble test`, without `verbose` flag.

image

genotrance:
- Show no command line and `--hints:off` output by default for calls to Nim
- Show command line and standard output for calls to Nim with `--verbose`

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -233,10 +233,16 @@ proc buildFromDir(
     raise newException(NimbleError,
         "Nothing to build. Did you specify a module to build using the" &
         " `bin` key in your .nimble file?")
-  var args = args
-  let nimblePkgVersion = "-d:NimblePkgVersion=" & pkgInfo.version
-  for path in paths: args.add("--path:\"" & path & "\" ")
-  var binariesBuilt = 0
+  var
+    binariesBuilt = 0
+    args = args
+  args.add "-d:NimblePkgVersion=" & pkgInfo.version
+  for path in paths:
+    args.add("--path:" & path.quoteShell)
+  if options.verbosity >= HighPriority:
+    # Hide Nim hints by default
+    args.add("--hints:off")
+
   let binToBuild =
     # Only build binaries specified by user if any, but only if top-level package,
     # dependencies should have every binary built.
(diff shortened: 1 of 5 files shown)


src/nimble.nim
```diff
@@ -233,10 +233,16 @@ proc buildFromDir(
...
-  let nimblePkgVersion = "-d:NimblePkgVersion=" & pkgInfo.version
-  for path in paths: args.add("--path:\"" & path & "\" ")
-  var binariesBuilt = 0
+  var
+    binariesBuilt = 0
+    args = args
+  args.add "-d:NimblePkgVersion=" & pkgInfo.version
+  for path in paths:
+    args.add("--path:" & path.quoteShell)
+  if options.verbosity == HighPriority:
```
dom96, line 242:
```suggestion
  if options.verbosity >= HighPriority:
```

src/nimble.nim
```diff
@@ -233,10 +233,16 @@ proc buildFromDir(
...
-  var args = args
-  let nimblePkgVersion = "-d:NimblePkgVersion=" & pkgInfo.version
-  for path in paths: args.add("--path:\"" & path & "\" ")
-  var binariesBuilt = 0
+  var
+    binariesBuilt = 0
+    args = args
+  args.add "-d:NimblePkgVersion=" & pkgInfo.version
+  for path in paths:
+    args.add("--path:" & path.quoteShell)
```
dom96, line 241: Nit: making these minor syntax changes just makes the git history harder to follow and IMO it was fine the way it was.

dom96:
Based on your other comments it seems that warnings/deprecations will still be shown. Why should the user see that when installing a package with hundreds of deps?

src/nimble.nim
genotrance: Done

genotrance:
> Based on your other comments it seems that warnings/deprecations will still be shown. Why should the user see that when installing a package with hundreds of deps?

One compromise is to only show the warnings/deprecations for the top level package but hide it for dependencies but this way it is arguable that developers will be subtly pressured to clean up their libraries.

Here's output with `--verbose`:

image

PR #829, opened 2020-07-28
Fix #757 - compile flags for test

### LINKED ISSUE #757: Pass compiler options to `nimble test`
I'm trying to pass options to the compiler when running the `nimble test`. I've tried the following:

via CLI:
```
nimble test --define:TEST_ITERATIONS=100
```

via overriding task:
```
task test, "Test Application.":
    switch("define", "TEST_ITERATIONS=100")
    setCommand "nop"
```

via delegation task:
```
task test2, "Test Application.":
    switch("define", "TEST_ITERATIONS=100")
    setCommand "test"
```

None of them propagated the options to `nim c ...`. The closest was the delegation task but it only passes the option `nim e` which makes sense (kind of).

genotrance:
- `nimble test` forwards non-nimble flags to compiler
- `--passNim` is no longer required for `nimble install` but still supported
- Pass compile flags to nimscript
- Cleaned up help
- Change log

### DIFF
--- a/changelog.markdown
+++ b/changelog.markdown
@@ -14,10 +14,10 @@ bug fixes with some minor improvements:
   compiler output.
 - Custom tasks can now be passed compiler flags as well as run flags when run
   as `nimble <compflags> task <runflags>`. This includes the custom `test`
-  task if defined. Compile flags can be used to set `--define:xxx` values that
-  can be checked with `when defined(xxx)` and other compiler flags that are
-  applicable in nimscript mode. Run flags can be accessed per usual from
-  `commandLineParams: seq[string]`.
+  task if defined. Compile flags are forwarded to `nim e` that executes the
+  `.nimble` task and can be used to set `--define:xxx` and other compiler flags
+  that are applicable in Nimscript mode. Run flags can be accessed per usual
+  from `commandLineParams: seq[string]`.
 - The default `nimble test` task also allows passing compiler flags but given
   run flags are not really applicable for multiple test binaries, it allows
   specifying compile flags before or after the `test` task.
--- a/changelog.markdown
+++ b/changelog.markdown
@@ -3,6 +3,39 @@
 
 # Nimble changelog
 
+## 0.11.6 - TBD
+
+This is a minor release containing just X commits. This release brings mostly
+bug fixes with some minor improvements:
+
+- `nimble dump` now provides `--json` output.
+- Calls to the Nim compiler now display `--hints:off` output by default.
+  The `--verbose` flag will print the full Nim command as well as regular
+  compiler output.
+- Custom tasks can now be passed compiler flags as well as run flags when run
+  as `nimble <compflags> task <runflags>`. This includes the custom `test`
+  task if defined. Compile flags can be used to set `--define:xxx` values that
+  can be checked with `when defined(xxx)` and other compiler flags that are
+  applicable in nimscript mode. Run flags can be accessed per usual from
+  `commandLineParams: seq[string]`.
+- The default `nimble test` task also allows passing compiler flags but given
+  run flags are not really applicable for multiple test binaries, it allows
+  specifying compile flags before or after the `test` task.
+- `nimble install` also allows passing compiler flags similar to the default
+  `nimble test` and no longer requires the `--passNim` flag.
+
+----
+
+Full changelog: https://github.com/nim-lang/nimble/compare/v0.11.4...master
+
+## 0.11.4 - 19/05/2020
+
+This is a minor release containing just 2 commits and a few minor bug fixes.
+
+----
+
+Full changelog: https://github.com/nim-lang/nimble/compare/v0.11.2...v0.11.4
+
 ## 0.11.2 - 02/05/2020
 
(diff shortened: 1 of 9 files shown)


src/nimblepkg/options.nim
```diff
@@ -386,10 +395,11 @@ proc parseFlag*(flag, val: string, result: var Options, kind = cmdLongOption) =
...
     result.setRunOptions(flag, getFlagString(kind, flag, val), false)
   of actionCustom:
-    if result.action.command.normalize == "test":
-      if f == "continue" or f == "c":
+    if result.action.command.normalize == "test" and
+      f == "continue" or f == "c":
         result.continueTestsOnFailure = true
-    result.action.flags[flag] = val
+    elif not isGlobalFlag:
+      result.setCustomOptions(getFlagString(kind, flag, val))
```
dom96, line 402: This means that a `custom` `test` task will not get a `--continue`/`-c` flag. I would consider that a bug.

src/nimblepkg/options.nim
```diff
@@ -67,16 +67,16 @@ type
     of actionCustom:
       command*: string
       arguments*: seq[string]
-      flags*: StringTableRef
+      flags*: seq[string]
```
dom96, line 70: Why the change to seq? A table is more flexible and supports ``--key:val``.

genotrance: Reusing the work done in `nimble run` as well as the `getFlagString()` code - there was no reason to store some flags as seqs and this one as a table.

src/nimblepkg/options.nim
genotrance: Why wouldn't it? It would get handled in the previous if condition right?

genotrance:
@dom96 - PTAL - this is ready to go.

src/nimblepkg/options.nim
dom96: The custom task won't be able to access the flags because they are not added to the "custom options". The only way they would be able to read it is if you make ``continueTestsOnFailure`` accessible. See what I mean?

src/nimblepkg/options.nim
dom96: Shouldn't they all be stored as tables to enable `--key:val`?

changelog.markdown
```diff
@@ -3,6 +3,39 @@
...
+- `nimble dump` now provides `--json` output.
+- Calls to the Nim compiler now display `--hints:off` output by default.
+  The `--verbose` flag will print the full Nim command as well as regular
+  compiler output.
+- Custom tasks can now be passed compiler flags as well as run flags when run
+  as `nimble <compflags> task <runflags>`. This includes the custom `test`
+  task if defined. Compile flags can be used to set `--define:xxx` values that
+  can be checked with `when defined(xxx)` and other compiler flags that are
+  applicable in nimscript mode. Run flags can be accessed per usual from
+  `commandLineParams: seq[string]`.
```
dom96, line 20: how are these compile flags and run flags used? Are they passed implicitly to any ``exec "nim .."`` calls done in the task? You should explain that explicitly.

changelog.markdown
```diff
@@ -3,6 +3,39 @@
...
+- The default `nimble test` task also allows passing compiler flags but given
+  run flags are not really applicable for multiple test binaries, it allows
+  specifying compile flags before or after the `test` task.
```
dom96, line 23: Might be worthwhile to support `nimble test <test_name>` to compile&run a specific test.

dom96:
Also, I'm a little concerned about just passing any unknown flag to Nim. Could we whitelist the flags instead of just passing any unknown flags to Nim?

src/nimblepkg/options.nim
genotrance: `--key:val` is supported by this method since it is used by `nimble run` as well.

src/nimblepkg/options.nim
genotrance: Ah you mean to forward that as a run flag in the new design. I'll clean it up accordingly.

changelog.markdown
genotrance: Not to the exec calls but to the `nim e` call that processes the task. I'll clarify.

User will have to use the run flags to manually pass stuff to the exec calls if they want.

src/nimblepkg/options.nim
genotrance: > Also, I'm a little concerned about just passing any unknown flag to Nim. Could we whitelist the flags instead of just passing any unknown flags to Nim?

Nim complains if it gets invalid flags. And I'd rather not maintain a list since Nim is a moving target of flags.

genotrance:
> Also, I'm a little concerned about just passing any unknown flag to Nim. Could we whitelist the flags instead of just passing any unknown flags to Nim?

Nim complains if it gets invalid flags. And I'd rather not maintain a list since Nim is a moving target of flags.

src/nimblepkg/options.nim
genotrance: Fixed.

changelog.markdown
genotrance: I'll leave this for another day.

PR #833, opened 2020-07-31
Remove dead code

stisa:
I think I found some dead code, lets see if the tests agree.
Looks like they were replace by the `iterXXX` family of procs in packageinfo?


stisa:
Looks like the failure was unrelated to the changes?
```
fatal: unable to access 'https://github.com/babel-test/packagea.git/': 
Failed to connect to github.com port 443: Operation timed out
```

PR #834, opened 2020-08-19
Local deps mode

genotrance:


src/nimble.nim
```diff
@@ -400,76 +400,81 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
...
+                  binariesInstalled)
+
+    # Save the nimble data (which might now contain reverse deps added in
+    # processDeps).
+    saveNimbleData(options)
+
+    # update package path to point to installed directory rather than the temp directory
+    pkgInfo.myPath = dest
+  else:
+    display("Warning:", "Skipped copy in local deps mode", Warning)
```
dom96, line 476: Nice!

src/nimble.nim
```diff
@@ -400,76 +400,81 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
...
-      let dest = changeRoot(realDir, pkgDestDir, file)
-      filesInstalled.incl copyFileD(file, dest)
-  )
-
-  # Copy the .nimble file.
-  let dest = changeRoot(pkgInfo.myPath.splitFile.dir, pkgDestDir,
-                        pkgInfo.myPath)
-  filesInstalled.incl copyFileD(pkgInfo.myPath, dest)
+  # Don't copy artifacts if local deps mode and "installing" the top level package
+  if not (options.localdeps and options.startDir == dir):
```
dom96, line 404: Suggestion: create a ``isInstallingTopLevel`` function (or similar name) which is implemented via ``options.startDir == dir``. Seems like it will be useful for the future (and will make the explanation tied to the code).

src/nimble.nim
```diff
@@ -1021,39 +1026,43 @@ proc developFromDir(dir: string, options: Options) =
...
+    # Save the nimble data (which might now contain reverse deps added in
+    # processDeps).
+    saveNimbleData(options)
 
-  display("Success:", (pkgInfo.name & " linked successfully to '$1'.") %
-          dir, Success, HighPriority)
+    display("Success:", (pkgInfo.name & " linked successfully to '$1'.") %
+            dir, Success, HighPriority)
+  else:
+    display("Warning:", "Skipping link in local deps mode", Warning)
```
dom96, line 1065: This should be an error, no?

src/nimble.nim
dom96: You're actually using the same code below, so it would already be used across two call sites, so I think that supports my suggestion.

dom96 (approved):
Looks good in general. See my comments though.

src/nimble.nim
genotrance: Similar to install, we are skipping creating a link of the top level project into its own deps directory.

src/nimble.nim
genotrance: This is done.

PR #839, opened 2020-09-01
Fix #835 - silence output

### LINKED ISSUE #835: Provide option to silence all non-essential output
Currently there's no way to silence non-essential output from the `nimble`, the code 

```
echo "Hi"
```

produces

```
$ nimble --hints:off run test
  Verifying dependencies for nim@1
      Info: Dependency on yaml@any version already satisfied
  Verifying dependencies for yaml@0.14.0
   Building nim/test using c backend
Hi
```

It's possible to silence nim with `nim c -r --hints:off test.nim` but not `nimble`

Would be nice to have an option so `nibmle` would produce only the program output, maybe something like `--quiet` so that the output will be

```
$ nimble --quiet run test
Hi
```

genotrance:
@dom96 - PTAL

### DIFF
--- a/changelog.markdown
+++ b/changelog.markdown
@@ -28,6 +28,7 @@ This is a major release containing multiple improvements and bug fixes:
   exists within a project, Nimble will use it to store all package dependencies
   instead of `~/.nimble/bin`. This enables isolation of a project and its
   dependencies from other projects being developed.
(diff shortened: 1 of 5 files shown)


dom96 (approved):
LGTM

PR #841, opened 2020-09-02
Fix #808 - relative symlinks

### LINKED ISSUE #808: Absolute paths in bin symlinks
`NIMBLE_DIR` can not be moved/renamed because the symlinks in `NIMBLE_DIR/bin` are absolute. Moving NIMBLE_DIR around is handy in CI environments.

genotrance:
cc @yglukhov

### DIFF
--- a/src/nimblepkg/packageinstaller.nim
+++ b/src/nimblepkg/packageinstaller.nim
@@ -30,7 +30,9 @@ when defined(windows):
 proc setupBinSymlink*(symlinkDest, symlinkFilename: string,
                       options: Options): seq[string] =
   result = @[]
-  let currentPerms = getFilePermissions(symlinkDest)
+  let
+    symlinkDestRel = relativePath(symlinkDest, symlinkFilename.parentDir())
+    currentPerms = getFilePermissions(symlinkDest)
   setFilePermissions(symlinkDest, currentPerms + {fpUserExec})
   when defined(unix):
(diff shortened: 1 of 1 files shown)


src/nimblepkg/packageinstaller.nim
```diff
@@ -62,14 +64,14 @@ proc setupBinSymlink*(symlinkDest, symlinkFilename: string,
       if fixChcp:
         contents.add "chcp 65001 > nul && "
       else: contents.add "chcp 65001 > nul\n@"
-    contents.add "\"" & symlinkDest & "\" %*\n"
+    contents.add "\"%~dp0\\" & symlinkDestRel & "\" %*\n"
```
dom96, line 67: damn, this is some magic. You might want to consider using raw string literals here.

dom96 (changes requested):
This seems good, but it needs a test plan. Can you show how you've tested this change together with the results?

genotrance:
> This seems good, but it needs a test plan. Can you show how you've tested this change together with the results?

I manually verified that the shims are created with relative paths on Win/Lin and the commands still work. The CI builds and runs many binaries so I figured it got tested as well.

src/nimblepkg/packageinstaller.nim
genotrance: Raw strings don't help since the string contains `"` and `\n`. Doesn't look any prettier.

PR #846, opened 2020-09-08
Fix #308, fix #515 - bin file name

### LINKED ISSUE #308: The pkg suffix rule
As described [here](https://github.com/nim-lang/nimble#hybrids):

> One thing to note about library and binary package hybrids is that your binary may share the name of the package. This will mean that you will not be able to put your .nim files in a pkgname directory. The reason you will not be able to do this is because binaries on some operating systems do not have an extension so they will clash with a directory of the same name.

> The current convention to get around this problem is to append pkg to the name as is done for nimble.

I've begun wondering whether this is the best solution. The other solution is to place the binaries in a different directory, but doing so has other issues. For example, if the binary depends on some external data that is stored inside the package directory, accessing it will not be easy.

### LINKED ISSUE #515: [Feature] Allow specifying name for the binary in binary packages
Currently the binary installed by binary packages always uses the same name as the Nim file. This is to limited, especially since file names in Nim must be valid Nim identifiers. I would like to use dashes in the binary name, which isn't possible right now.

genotrance:


src/nimblepkg/tools.nim
```diff
@@ -164,3 +164,12 @@ proc getNimbleUserTempDir*(): string =
...
   return tmpdir
+
+proc getSymlinkFilename*(filename: string, options: Options): string =
+  let spl = filename.splitFile
+  result = options.getBinDir() / (block:
+    if spl.ext == ".out":
+      spl.name
+    else:
+      spl.name.addFileExt(spl.ext)
+  )
```
dom96, line 175: Doesn't this work without the `block`?

dom96: if not just refactor the `if` into a variable above this line.

src/nimblepkg/packageparser.nim
```diff
@@ -258,10 +258,16 @@ proc readPackageInfoFromNimble(path: string; result: var PackageInfo) =
             for i in ev.value.multiSplit:
               if i.splitFile().ext == ".nim":
                 raise newException(NimbleError, "`bin` entry should not be a source file: " & i)
+              let (binSrc, bin) = if '=' notin i: (i, i) else:
+                let spl = i.split('=', 1)
+                (spl[0], spl[1])
```
dom96, line 263: I don't like this. I think you should introduce a new field called `namedBin` or `aliasedBin` or something and use Nim's tuple syntax. Don't make up even more arbitrary parsing rules(!)

src/nimblepkg/packageparser.nim
```diff
@@ -258,10 +258,16 @@ proc readPackageInfoFromNimble(path: string; result: var PackageInfo) =
...
               if result.backend == "js":
-                result.bin.add(i.addFileExt(".js"))
+                result.bin.add(bin.addFileExt(".js"))
               else:
-                result.bin.add(i.addFileExt(ExeExt))
+                result.bin.add(bin.addFileExt(ExeExt))
+              if dirExists(result.binDir / result.bin[^1]):
+                result.bin[^1] &= ".out"
```
dom96, line 269: huh? Why this weird code?

```suggestion
                result.bin.add(".out")
```

src/nimblepkg/common.nim
```diff
@@ -38,6 +38,7 @@ when not defined(nimscript):
       installExt*: seq[string]
       requires*: seq[PkgTuple]
       bin*: seq[string]
+      binSrc*: seq[string]
```
dom96, line 41:
```suggestion
      binSrc*: seq[string]
      binName*: seq[string]
```

dom96: or maybe even 
```suggestion
      bin*: seq[(string, string)]
```

dom96 (changes requested):

src/nimblepkg/packageparser.nim
genotrance: Basically issue #308, if binary file has same name as a directory, it won't get installed so we add a `.out` to the binary name.

src/nimblepkg/tools.nim
genotrance: Done.

src/nimblepkg/common.nim
genotrance: This will cascade into many code changes - everywhere `bin` is referenced.

src/nimblepkg/packageparser.nim
genotrance: I don't like it either but did this as an example - it results in minimal code changes. Adding a new field will be more code both in implementation (need to get the data from `nim e` back into nimble) as well as for users.

```nim
bin = @["test1"]
namedBin = @[("test2", "test-2")]
```

vs.

```nim
bin = @["test1", "test2=test-2"]
```

There's also the following example if we use tables:

```nim
bin = @["test1"]
namedBin["test2"] = "test-2"
```

genotrance: Okay, I could get `namedBin: Table[string, string]` for really cheap - data is still forwarded as `test2=test-2` in `bin` but the user doesn't need to see that. They can use either of:

```nim
namedBin["test2"] = "test-2"
namedBin = {"test2": "test-2"}.toTable
```

src/nimblepkg/packageparser.nim
genotrance: This also got cleaned up in the update. I still want to improve the `.out` handling cause `nimble run a` doesn't work if it is compiled as `a.out`.

src/nimblepkg/common.nim
genotrance: Was able to minimize the changes with a table.

genotrance:
@dom96 - PTAL - ready to go.

src/nimblepkg/packageinfo.nim
```diff
@@ -413,6 +413,8 @@ proc getOutputDir*(pkgInfo: PackageInfo, bin: string): string =
     result = pkgInfo.mypath.splitFile.dir / pkgInfo.binDir / bin
   else:
     result = pkgInfo.mypath.splitFile.dir / bin
+  if bin.len != 0 and dirExists(result):
+    result &= ".out"
```
dom96, line 417: Nit; but I prefer `add` over the `&=` it's far more idiomatic Nim

src/nimblepkg/nimscriptapi.nim
```diff
@@ -106,6 +105,11 @@ proc printPkgInfo(): string =
   if backend.len == 0:
     backend = "c"
 
+  # Forward `namedBin` entries in `bin`
+  for k, v in namedBin:
+    if k notin bin:
+      bin.add k & "=" & v
```
dom96, line 111: This still sucks and means you need to have code for parsing this new format (plus I'm sure this will have edge cases which will need to be covered in the future...)

Why can't namedBin be the source of truth?

dom96 (changes requested):
I like the idea to use a table, but I don't like that there is still this `=` parsing

src/nimblepkg/nimscriptapi.nim
genotrance: All said and done, we have to send the data from nimscript to nimble in an ini format.

https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/nimscriptapi.nim#L101

No matter what, we will need some `a = b` equivalent formatting to pull this off.

Issue #849, opened 2020-09-12
Nimble test does not walk subdirectories


TomDobbelaere:
Why?

PR #850, opened 2020-09-12
Use walkDirRec for nimble test, output to build/tests

TomDobbelaere:
This is not meant to be added into master directly, there's probably a lot wrong with this code style-wise but it works for me.

It serves as an illustration to this issue: #849 

This is how I expected `nimble test` to actually work, not whatever it is doing now.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -332,9 +332,9 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
     var filesInstalled = initHashSet[string]()
     iterInstallFiles(realDir, pkgInfo, options,
       proc (file: string) =
-        createDir(changeRoot(realDir, pkgDestDir, file.splitFile.dir))
-        let dest = changeRoot(realDir, pkgDestDir, file)
-        filesInstalled.incl copyFileD(file, dest)
+      createDir(changeRoot(realDir, pkgDestDir, file.splitFile.dir))
+      let dest = changeRoot(realDir, pkgDestDir, file)
+      filesInstalled.incl copyFileD(file, dest)
     )
 
     # Copy the .nimble file.
@@ -356,7 +356,8 @@ proc installFromDir(dir: string, requestedVer: VersionRange, options: Options,
           else: bin
         if fileExists(pkgDestDir / binDest):
           display("Warning:", ("Binary '$1' was already installed from source" &
-                              " directory. Will be overwritten.") % bin, Warning,
+                              " directory. Will be overwritten.") % bin,
+                              Warning,
                   MediumPriority)
 
         # Copy the binary file.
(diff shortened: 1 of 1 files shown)


genotrance:
CI is failing since this functionality is explicitly disabled - see `tests/testCommand/testsIgnore`.

Best to wait for @dom96 to confirm why this is not supported. It might very well be a breaking change but seems like a reasonable ask as well.

TomDobbelaere:
Thanks, don't mean to be a bother. I was just starting a project in Nim and this particular thing just bothered me is all.

src/nimble.nim
```diff
@@ -1025,23 +1027,23 @@ proc test(options: Options) =
...
       optsCopy.action = Action(typ: actionCompile)
-      optsCopy.action.file = file.path
+      optsCopy.action.file = file
       optsCopy.action.backend = pkgInfo.backend
       optsCopy.getCompilationFlags() = options.getCompilationFlags()
       # treat run flags as compile for default test task
       optsCopy.getCompilationFlags().add(options.action.custRunFlags)
       optsCopy.getCompilationFlags().add("-r")
+      optsCopy.getCompilationFlags().add("--out=" & ("build/" /
+          file.changeFileExt(ExeExt)))
```
dom96, line 1045: hm, don't we have a setting for this nowadays? @genotrance?

If not, maybe it would be better to put this in `tests/build/` or `tests/bin/`?

dom96 (commented):
Seems fine to me in principle, but please revert the code style changes, most are wrong and they make review harder.

src/nimble.nim
genotrance: Not sure what setting?

Related: https://github.com/nim-lang/nimble/issues/787

I think we should instead rely on `nim r` instead of worrying about yet another directory. That's only in devel though so we should wait a bit until it is in stable.

TomDobbelaere: My problem with relying on `nim r` is that it can leave executable files littered in folders, since in the original code you have to do a 'existedBefore' and 'existedAfter' check. If executable files, due to some crash or whatever reason, were not cleaned up properly, you now have these files scattered all over the test folder's subdirectories that will -not- get removed next 'nimble test' run (because existedBefore will now be true, so they won't get trashed).

Traditionally, a build folder is very clear that it's just build output and can be safely deleted.

I guess `tests/build` could work. 

I just think it's important to have this discussion because it would make `nimble test` plug and play.

My workflow is to mirror my 'tests' folder based on my src folder, more or less.
So if I had a `src/foo/foo_database.nim`, it would have a corresponding `tests/foo/test_foo_database.nim`

dom96:
What's the status of this? Can we get those formatting changes removed?

dom96:
ping

PR #851, opened 2020-09-12
Perf improvements

### LINKED ISSUE #830: nimble init at odds with documentation
Wanting to start a new library package, one issues `nimble init` from their project root directory. The following structure is produced:

```
├── src
│   ├── mylib
│   │   └── submodule.nim
│   └── mylib.nim
├── test
│   ├── config.nims
│   └── test1.nim
└── mylib.nimble
```

But [the docs](https://github.com/nim-lang/nimble#creating-packages) say "If your package exposes only a single module, then that module should be present in the root directory (the directory with the .nimble file) of your Git repository, and should be named whatever your package's name is."

Maybe changing init's behaviour by prompting the user if they want a multi-module or single?

genotrance:
- Load packages.json only once
- Fix #830

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -627,11 +627,10 @@ determined by the nature of your package, that is, whether your package exposes
 only one module or multiple modules.
 
 If your package exposes only a single module, then that module should be
-present in the root directory (the directory with the .nimble file) of your Git
-repository, and should be named whatever your package's name is. A good example
-of this is the [jester](https://github.com/dom96/jester) package which exposes
-the ``jester`` module. In this case the jester package is imported with
-``import jester``.
+present in the source directory of your Git repository, and should be named
+whatever your package's name is. A good example of this is the
+[jester](https://github.com/dom96/jester) package which exposes the ``jester``
+module. In this case the jester package is imported with ``import jester``.
 
 If your package exposes multiple modules then the modules should be in a
 ``PackageName`` directory. This will allow for a certain measure of isolation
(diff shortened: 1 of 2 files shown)


genotrance:
@dom96 - PTAL - not thrilled to use a global but `options` is not writable either.

dom96 (approved):
Looks fine. Only as long as the global is tied to the memoization of a proc.

I've been wanting a `{.memoize.}` pragma for a while which implements this, perhaps you could create a PR for one into the Nim stdlib?

genotrance:
There's the [memo](https://github.com/andreaferretti/memo) package - I tested it and it works fine - we can simply copy memo.nim into nimble if you are good with that.

I didn't see any other places where this would help much within nimble with the profiling I did.

dom96:
Right... I think it would be nice to suggest the inclusion of this in the stdlib. It's simple enough and small enough and useful enough to be worthwhile.

I guess we can include it in Nimble for now. Perhaps we should just use submodules for this?

Btw if we do include it in the stdlib, I'd like `{.memoized.}` to be renamed to `{.memoize.}` mainly because that's what I'm used to in other langs :P

genotrance:
Using submodules might drive changes into koch. I think I'll leave this as is for now and push for memoize to get added to the stdlib in the interim. Meanwhile, on the nimble side, if I see other instances where memoize will benefit us, will include it and remove this global.

dom96:
Sure, this also ties into the generic future possibility of us depending on Nimble packages in Nimble. It will hopefully become more and more common that we may wish to do so.

PR #852, opened 2020-09-14
Localdeps flag support

genotrance:


src/nimblepkg/options.nim
```diff
@@ -130,6 +134,7 @@ Nimble Options:
   -v, --version                   Print version information.
   -y, --accept                    Accept all interactive prompts.
   -n, --reject                    Reject all interactive prompts.
+  -l, --local                     Run in project local dependency mode
```
dom96, line 137: I would still vote for ``localdeps``, it's better for it to be explicit and you have a short flag too.

dom96 (approved):
In general looks good.

src/nimblepkg/options.nim
genotrance: Done

PR #855, opened 2020-09-25
Improve compilation messages

cdunn2001:
* Show path when `nimble` is not found, for debugging.
* Suppress ObservableStores warning.

### DIFF
--- a/nim.cfg
+++ b/nim.cfg
@@ -1,0 +1,3 @@
+@if nimHasWarningObservableStores:
+  --warning[ObservableStores]: off
+@end
--- a/tests/tester.nim
+++ b/tests/tester.nim
@@ -20,7 +20,7 @@ removeDir(installDir)
 createDir(installDir)
 
 # Always recompile.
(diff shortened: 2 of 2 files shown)


nim.cfg
```diff
@@ -0,0 +1,3 @@
+@if nimHasWarningObservableStores:
+  --warning[ObservableStores]: off
```
dom96, line 2: https://forum.nim-lang.org/t/6442#39738

> There is a new warning about "Observable Stores". I have yet to see a case where this warning indicates a real problem and so your config should probably disable it. Code to do that is:

lol? Why did you implement this warning and enable it by default then @Araq?

I'd honestly rather not pollute Nimble's nim.cfg with stuff like this if the compiler appears to be to blame.

cdunn2001: Yeah, but for now it occurs soooo many times. It dominates the output.

dom96: I'd rather wait for the warning to be disabled by default which it sounds like it should be.

dom96:
I hope that this was resolved in Nim, but let's discuss it in that repo if not.

PR #856, opened 2020-10-06
Test(dump): Add tests for `--ini` and `--json`

ee7:
Previously, there wasn't a test for the `--ini` or `--json` switches (which were added by #810).

This PR also adds tests for the full `nimble dump` output. Previously, we just tested that there's a line that contains:
`"desc: \"Test package for dump command\""`


dom96 (approved):
Epic, thanks!

PR #857, opened 2020-10-06
cli(help): Document `nimble dump [--ini, --json]`

ee7:
Is the `--json` switch considered stable enough to document? Feel free to suggest changes to the wording.

`nimble dump --json` was added in 9a3c569f6c4e (#810), but was
not shown when running `nimble --help`.

This commit also documents the fact that the `--ini` option exists, and
that INI is the default output format.


dom96 (approved):
LGTM

PR #862, opened 2020-10-14
patch version can be a multi-digit number

narimiran:
This is crucial for Nim 1.0.10 release, because currently 1.0.10 doesn't satisfy `>= 1.0.4` requirement.

cc @genotrance

### DIFF
--- a/src/nimblepkg/tools.nim
+++ b/src/nimblepkg/tools.nim
@@ -57,7 +57,7 @@ template cd*(dir: string, body: untyped) =
 proc getNimrodVersion*(options: Options): Version =
   let vOutput = doCmdEx(getNimBin(options).quoteShell & " -v").output
   var matches: array[0..MaxSubpatterns, string]
-  if vOutput.find(peg"'Version'\s{(\d+\.)+\d}", matches) == -1:
+  if vOutput.find(peg"'Version'\s{(\d+\.)+\d+}", matches) == -1:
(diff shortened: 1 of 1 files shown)


genotrance:
1.0.x has an old version of nimble - do we want to create a branch for this?

narimiran:
> 1.0.x has an old version of nimble - do we want to create a branch for this?

Then we should create a branch for 1.0.x and a branch for 1.2.x too, because this same problem will hit us if/when we release 1.2.10.

Here are commits used currently, in hope it makes easier for you to create branches:
* version-1-0: 4007b2a778429a978e12307bf13a038029b4c4d9
* version-1-2: 63695f490728e3935692c29f3d71944d83bb1e83

dom96:
Instead of creating branches we should just ship latest Nimble. No need for this extra work, the extra Nimble changes are unlikely to cause problems. If they do we can release Nim 1.0.12 ;)

genotrance:
cc @Araq for his thoughts on this - much easier to do this and it is ready to go (backport https://github.com/nim-lang/Nim/pull/15573 to older versions).

narimiran:
I just spoke with @Araq and he agrees: we will ship the latest nimble with all three Nim versions. (I'll backport https://github.com/nim-lang/Nim/pull/15573)

Araq:
> Instead of creating branches we should just ship latest Nimble.

Yes, I completely agree.

PR #868, opened 2020-10-29
Enable GitHub actions

iffy:
This builds and tests Nimble on the nightly builds of these versions:

- 1.0.x
- 1.4.x
- devel

It lacks building for 0.20.x

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - uses: alaviss/setup-nim@master
+      - uses: alaviss/setup-nim@0.1.0
         with:
           path: 'nim'
           version: ${{ matrix.nim }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,7 @@ jobs:
         with:
           path: 'nim'
           version: ${{ matrix.nim }}
-      - run: echo $PATH
       - run: nim --version
-      - run: pwd
       - run: |
           cd tests
           nim c -r tester
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: test
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        os: ['windows-latest', 'macos-latest', 'ubuntu-latest']
+        nim: ['devel', 'version-1-4', 'version-1-0']
+
+    name: Nim ${{ matrix.nim }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: alaviss/setup-nim@master
+        with:
+          path: 'nim'
+          version: ${{ matrix.nim }}
+      - run: |
+          cd tests
+          nim c -r tester
+      - run: ./src/nimble install -y
+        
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+    strategy:
+      matrix:
+        os: ['windows-latest', 'macos-latest', 'ubuntu-latest']
+        nim: ['devel', 'version-1-4', 'version-1-0']
+
+    name: Nim ${{ matrix.nim }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: alaviss/setup-nim@master
```
dom96, line 18: What's `master` mean here? Nim @ `master` branch? Can we pin to a known version or a commit like we do in the travis file?

dom96 (changes requested):

.github/workflows/ci.yml
iffy: It means use the master branch of https://github.com/alaviss/setup-nim (which is used to install nim). It doesn't refer to the master branch of Nim.  As written currently, this builds against Nim 1.0.x, Nim 1.4.x and Nim devel.  I'm not sure if alaviss/setup-nim can install a specific commit.

genotrance:
Apparently we need to merge this to a branch to test it - else it won't run.

@alaviss

.github/workflows/ci.yml
alaviss: Please use the 0.1.0 tag for stability. I'm planning a major overhaul in 0.2.0 which will break things.

iffy:
@genotrance you can see the actions run on my fork (from which this PR is made): https://github.com/iffy/nimble/actions/runs/339721125

.github/workflows/ci.yml
iffy: @alaviss Done. Thanks for the heads up!

dom96: Can we remove ``devel`` for now then? Having failures because we depend on a moving target is a source of frustration that I don't want to put on myself, or our contributors.

@alaviss I would appreciate if you added support for commits.

iffy: @dom96 Which commit would you like for now?  It has to be one that has a nightly build.  Unfortunately, the one in travis (8a004e2) doesn't have one.  It looks like that was from April of this year.  If you want one from back then, this looks like a good candidate (though it lacks macOS binaries): https://github.com/nim-lang/nightlies/releases/tag/2020-04-20-devel-1bdc30b

I'd recommend a more recent one, though.  The builds pass on this version: https://github.com/nim-lang/nightlies/releases/tag/2020-11-03-devel-fde17b159fdc96eda5609fd98e1fe3c7a34ef40b

I've hacked together something that will build a specific commit until alaviss/setup-nim supports building from arbitrary (or semi-arbitrary) commits.

alaviss: @dom96 It will be supported in the next version (which will be written in Nim). The current one focus on `fusion` needs, and we don't pin commits there.

dom96: > which will be written in Nim targeting NodeJS

huh? Why targeting Node?

dom96 (approved):
I remember the days when setting up Nim in CI was as simple as downloading a tarball and running ``sh build.sh && nim c -r koch boot -d:release``.

I'll approve but I urge everyone to consider whether so much complexity is warranted.

iffy:
Maybe https://github.com/nim-lang/nimble/pull/871 would be better?

iffy:
Superseded by #877

PR #870, opened 2020-11-09
export readPackageInfoFromNimble for external usage

bung87:


bung87:
it will not work for external usage. close it.

PR #871, opened 2020-11-09
Simpler GitHub Actions CI

iffy:
This enables GitHub Actions like #868 but instead of using pre-built nightlies, builds from source.  Here are some differences from the other PR:

- This CI code is simpler
- This CI code is not based on any third party repos
- This CI runs slower (since it builds Nim). This build takes ~14min for everything while #868 takes ~8min

Here is the latest CI run on my fork: https://github.com/iffy/nimble/actions/runs/354442960

Thank you, @dom96 for making me stop to think if the complexity of the other approach was worth it.

PR #877, opened 2020-11-18
Add Github Actions CI using choosenim

iffy:
This is yet another alternative for switching from Travis to Github Actions.  It supersedes #871 and #868 with the following differences:

- It uses choosenim to test Nim versions 0.20.2, 1.0.10, 1.2.8 and 1.4.0 (so as choosenim improves, so does this)
- It also tests with Nim built from Git SHA 8a004e2
- It also tests with this nightly https://github.com/nim-lang/nightlies/releases/tag/2020-11-11-devel-bbe49a14ae827b6474d692042406716a3b3dd71f
- @genotrance requested "I would prefer one place where all this is solved, not yet another copy of code that has to be maintained in each repo." This does that by using the Github Action [iffy/install-nim@v3](https://github.com/iffy/install-nim) instead of including a script file.

You can see the build run for this PR here: https://github.com/iffy/nimble/actions/runs/370324359

I apologize for making 3 PRs for this issue, but they haven't followed a sequential pattern of development.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,6 @@ jobs:
           - 1.0.10
           - 1.2.8
           - 1.4.0
-          - git:8a004e2
           - nightly:https://github.com/nim-lang/nightlies/releases/tag/2020-11-11-devel-bbe49a14ae827b6474d692042406716a3b3dd71f
     name: ${{ matrix.os }} - ${{ matrix.target }}
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,17 +10,19 @@ jobs:
           - windows-latest
           - macos-latest
           - ubuntu-latest
-        target:
+        nimversion:
           - 0.20.2
           - 1.0.10
           - 1.2.8
           - 1.4.0
           - nightly:https://github.com/nim-lang/nightlies/releases/tag/2020-11-11-devel-bbe49a14ae827b6474d692042406716a3b3dd71f
-    name: ${{ matrix.os }} - ${{ matrix.target }}
+    name: ${{ matrix.os }} - ${{ matrix.nimversion }}
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - uses: iffy/install-nim@v3
+        with:
+          version: ${{ matrix.nimversion }}
       - run: nim --version
       - name: Run nim c -r tester
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+      matrix:
+        os:
+          - windows-latest
+          - macos-latest
+          - ubuntu-latest
+        target:
+          - 0.20.2
+          - 1.0.10
+          - 1.2.8
+          - 1.4.0
```
dom96, line 17: IMO we only need these:

```suggestion
          - 1.0.10
          - 1.4.0
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+          - git:8a004e2
```
dom96, line 18: is this just a demo? :) 

Pretty cool, but we don't need this here.

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+          - nightly:https://github.com/nim-lang/nightlies/releases/tag/2020-11-11-devel-bbe49a14ae827b6474d692042406716a3b3dd71f
```
dom96, line 19: I'm guessing choosenim is not used for this case?

dom96 (commented):
Cool. Do I dare look at the implementation of iffy/install-nim? :)

iffy:
> Cool. Do I dare look at the implementation of iffy/install-nim? :)

Nope, you'd better not :)  The car runs -- don't pop the hood.

.github/workflows/test.yml
iffy: This mimics the current Travis build: https://github.com/nim-lang/nimble/blob/master/.travis.yml#L13 I'm fine removing it.

.github/workflows/test.yml
iffy: Correct. If there's a `something:` prefix, choosenim is not used.

.github/workflows/test.yml
dom96: well, it's either this or the nightly. We don't need both. If the nightly works and is newer let's keep it instead of this.

dom96 (approved):

iffy:
Don't merge yet -- nightly isn't working. I'll fix it.

iffy:
Okay, now nightly is building right: https://github.com/iffy/nimble/actions/runs/373098791

dom96:
Epic, thanks!

PR #882, opened 2020-12-20
Update packageparser.nim

lihf8515:
Blank lines and comment lines in 'parsecfg' are ignored when read.
I'm going to add the ability to keep blank lines and comment lines to 'parsecfg'.

 'key' starting with 'cfgBlankAndCommentLine' is the blank line and comment line read by the 'parsecfg' module. An exception should not be thrown; it should be skipped.

[parsecfg # 16402](https://github.com/nim-lang/Nim/pull/16402)

### DIFF
--- a/src/nimblepkg/packageparser.nim
+++ b/src/nimblepkg/packageparser.nim
@@ -291,6 +291,8 @@ proc readPackageInfoFromNimble(path: string; result: var PackageInfo) =
           else:
             if not startsWith(ev.key, "cfgBlankAndCommentLine"):
               raise newException(NimbleError, "Invalid field: " & ev.key)
+        of "":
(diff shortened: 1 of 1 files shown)


lihf8515:
Why are they always automatically cancelled?

dom96 (changes requested):
This code is legacy, Nimble reads via nimscript nowadays so we shouldn't really change it. I also don't see how this change is valuable.

lihf8515:
I submitted code to "parsecfg" to retain comment lines. But the CI test failed. The reason is that "nimble" code is used in CI test code, and "nimble" does not deal with comment lines, so I need to modify "nimble" code, but it does not affect any function of "nimble".

PR #886, opened 2021-01-10
Always do shallow copies of submodules with depth 1

greenfork:
For large libraries such as https://github.com/raysan5/raylib it can go from 6 minutes to 30 seconds with this fix


dom96 (approved):
Seems good. My only concern is regarding the `--shallow-submodules` flag. I couldn't find when it was implemented in git but hopefully it is implemented in fairly old git versions (we've had people run into problems because they had a too old git version in the past).

greenfork:
@dom96 its was added on Apr 26, 2016 in this commit: https://github.com/git/git/commit/d22eb044750631f30bdbfc3dc42d684e73d18ddc

yglukhov:
Unfortunately this is pretty fragile, and because of it packages can "break" over time. See example: https://github.com/status-im/nim-bearssl/issues/16

PR #891, opened 2021-02-05
Bump iffy/install-nim to v3.1 for increased logging

iffy:
This adds additional logging to the GitHub actions that install from nightlies in an attempt to figure out what happened to cause https://github.com/iffy/install-nim/issues/8

I suspect the issue is transient (e.g. the request to the GitHub API failed for some reason)


dom96:
Nice, this gives more info.

image

PR #459, opened 2018-02-18
Add SSL certificate check handling

FedericoCeratto:
Depends on https://github.com/nim-lang/Nim/pull/6664

PR #893, opened 2021-02-05
Restore CVerifyPeer by default.

dom96:


FedericoCeratto:
Basic testing described on #459

dom96:
Thanks, but it doesn't work for cases that have valid SSL like GitHub which breaks Nimble completely since it cannot download its package list.

PR #896, opened 2021-02-07
Change option parsing for `nimble run`

### LINKED ISSUE #889: Compiler options for `nimble run`
I did not find a way to pass compiler options (like `--assertions:off` for example) to `nimble run`, instead the options get passed to the executable.

`nimble --help` says the following;
```
Builds and runs a package.
Binary needs to be specified after any
compilation options if there are several
binaries defined. Any flags after the binary
or -- arg are passed to the binary when it is run.
```
This implies to me that `nimble run <compiler options> <executable name>` or `nimble run <compiler options> --` should work, but they don't.

Is there anything I missed, that allows me to directly pass compiler options via `nimble run`? Or is this a bug?

konsumlamm:
Fixes #889.

This changes the option parsing for `nimble run` to use the flags before the executable name or `--` as compiler flags. For example, `nimble run --assertions:off -- arg` now parses `--assertions:off` as compiler flag. As mentioned in the linked issue, with this change, `nimble run <arguments>` (where the arguments are supposed to be passed to the program) wouldn't work anymore, it would need to be changed to `nimble run -- <arguments>`.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -1,4 +1,4 @@
-# Nimble 
+# Nimble
 
 Nimble is a *beta*-grade *package manager* for the [Nim programming
 language](https://nim-lang.org).
@@ -124,7 +124,7 @@ to obtain a list of available commands.
 The ``refresh`` command is used to fetch and update the list of Nimble packages
 (see below). There is no automatic update mechanism, so you need to run this
 yourself if you need to *refresh* your local list of known available Nimble
-packages.  Example:
+packages. Example:
 
     $ nimble refresh
     Downloading package list from https://.../packages.json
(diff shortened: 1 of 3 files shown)


dom96 (commented):
AFAIK Nim doesn't support the `--` concept, or has that changed? Can we be consistent with Nim?

konsumlamm:
> AFAIK Nim doesn't support the `--` concept, or has that changed? Can we be consistent with Nim?

`--` doesn't exist for `nim`, because it makes no sense there. The purpose of `--` is that you don't have to type out the executable name when there only is one available (`nimble run <options> -- <arguments>` is the same as `nimble run <options> <name> <arguments>`, where `<name>` is the name of the single executable, this just wasn't necessary before, since all arguments after the `run` were passed to the executable). Note that I did not add support for `--` here, that already existed, so I don't get why you want to remove it. It doesn't make sense to remove it "for consistency with `nim`" IMO, because the use case is a bit different.

dom96 (approved):
Ahh, makes sense. Thanks for the explanation.

PR #902, opened 2021-02-23
Add GITHUB_TOKEN to avoid rate limits

iffy:
Fixes https://github.com/iffy/install-nim/issues/8

If you'd like to review what changed between v3.1 and v3.2.0 of install-nim, here's the diff: https://github.com/iffy/install-nim/compare/v3.1...v3.2.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,9 +16,11 @@ jobs:
           - nightly:https://github.com/nim-lang/nightlies/releases/tag/2021-02-06-devel-39230422d02bd41b02378211e8613f702e5b945c
     name: ${{ matrix.os }} - ${{ matrix.nimversion }}
     runs-on: ${{ matrix.os }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v2
-      - uses: iffy/install-nim@v3.1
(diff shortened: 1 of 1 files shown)


Araq:
CIs is completely red with this change. Strange.

iffy:
I don't understand why the tests are failing. It looks like only one test ran (the others were canceled at the first failure): https://github.com/nim-lang/nimble/pull/902/checks?check_run_id=1963277877 And the failure is for something that happens after install-nim is done.

1. Would you like all tests to continue running even if one test fails? I can make that change if you'd like.
2. Maybe the tests behave differently when `GITHUB_TOKEN` is set?

dom96:
Well, it was broken in https://github.com/nim-lang/nimble/commit/d13f3b8ce288b4dc8c34c219a4e050aaeaf43fc9

It's clearly caused by this change, here is the error:

> /home/runner/work/nimble/nimble/tests/nimbleDir/pkgs/nimble-#head/nimblepkg/common.nim(70, 21) Error: cannot open file: ../../nimble.nimble

I know it's annoying having to update the version in all 2 files... but at least it doesn't lead to annoying failures in CI.

Araq:
> I know it's annoying having to update the version in all 2 files... but at least it doesn't lead to annoying failures in CI. Can we revert this change?

Not if I can help it, I hate having to update the version in two places.

Issue #907, opened 2021-03-18
[Enhancement request]: Stick to patchable version: requires ^= 0.1.0


marcomq:
Currently, nimble only supports `requirements >= version` and `requirements == version` according to the documentation.

Python has some easy requirements.txt [options ](https://pip.pypa.io/en/latest/reference/pip_install/#example-requirements-file) for pip, while node/npm packages.json has some even more complex [syntax](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#dependencies). 

Usually, when trying out python libraries or examples, the most used seems to be the ` == version` operator but there is also a less used but available `~=` operator that only installs patches.
For node/npm, tjhe default seems to be a `^` prefix, which always installs the latest patch/minor version. A  `~` is also available in node but not used that much.
There is an automatic `npm update` command to automatically change the `packages.json`  to the latest version of packages, even for those with `^`  or `~` prefix.

For nim, there is mostly a `>=` operator available, and there is also a new `==` operator for nim versions >= 1.2.
I also found a Nim syntax for requirements in the latest code `>= 1.5 & <=1.8` in /src/nimblepkg/version.nim:parseVersionRange that also seems to work but it is undocumented yet.

I would prefer a simpler npm like syntax:  `^= 1.1.1` which may be converted internally to `>= 1.1.1 & < 2.0.0`  or  `~= 1.1.1` which might be converted to `>= 1.1.1 & < 1.2.0` 
I haven't seen `&` in nim yet with `and` meaning and this actually seems to be a bit complicate for everyday usage.

dom96:
sure, AFAIK we've got other issues for this, but I'll keep this open. It would be nice to get some sugar for an operator that means "Give me the semver compatible version"

marcomq:
Cool. I will add a PR for this.

dom96:
Fixed via #908.

PR #908, opened 2021-03-19
Add ~ and ^ semver semantic to nimble "requires"

marcomq:
Added ~ and ^ semver semantic to nimble "requires" 
Check https://devhints.io/semver for semver  examples
#907

### DIFF
--- a/src/nimblepkg/version.nim
+++ b/src/nimblepkg/version.nim
@@ -154,11 +154,12 @@ proc makeRange*(version: string, op: string): VersionRange =
   result.ver = Version(version)
 
 proc parseVersionStringAt(s: string, i: var int): string = 
+  ## parse version string after ~ or ^ and return it
   while i < s.len:
     case s[i]
     of '0'..'9', '.':
       result.add(s[i])
-    of 'x', 'X', '*':
+    of 'x', 'X', '*': # only allowed for ~ and ^
       result.add("0")
     of '=':
       discard
@@ -173,7 +174,8 @@ proc parseVersionStringAt(s: string, i: var int): string =
           "Unexpected char in version range '" & s & "': " & s[i])
     inc(i)
 
-proc getNextExcludedVersion(version: string, nextMajor: bool): string = 
+proc getNextExcludedVersion(version: string, majorCompatibility: bool): string = 
+  ## try to get next higher version to exclude according to semver semantic
   var numbers = version.split('.')
   while numbers.len < 3:
     numbers.add("0")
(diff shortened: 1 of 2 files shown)


src/nimblepkg/version.nim
```diff
@@ -316,12 +366,33 @@ when isMainModule:
...
   doAssert findLatest(parseVersionRange(">= 0.1 & <= 0.4"), versions) ==
       (newVersion("0.2.3"), "v0.2.3")
+  doAssert findLatest(parseVersionRange("^= 0.1"), versions) ==
+      (newVersion("0.1.1"), "v0.1.1")
+  doAssert findLatest(parseVersionRange("^= 0.0.1"), versions) ==
+      (newVersion("0.0.1"), "v0.0.1")
+  doAssert findLatest(parseVersionRange("^= 2.2.2"), versions) ==
+      (newVersion("2.3.2"), "v2.3.2")
+  doAssert findLatest(parseVersionRange("^2.2.2"), versions) ==
+      (newVersion("2.3.2"), "v2.3.2")
```
dom96, line 391: Hmm, nit but I think I'd rather enforce the need to write `^=`

src/nimblepkg/version.nim
```diff
@@ -153,6 +153,46 @@ proc makeRange*(version: string, op: string): VersionRange =
...
   result.ver = Version(version)
 
+proc parseVersionStringAt(s: string, i: var int): string = 
+  ## parse version string after ~ or ^ and return it
+  while i < s.len:
+    case s[i]
+    of '0'..'9', '.':
+      result.add(s[i])
+    of 'x', 'X', '*': # only allowed for ~ and ^
+      result.add("0")
```
dom96, line 163:
```suggestion
    of '*': # only allowed for ~ and ^
      result.add("0")
```

Let's only support one variant, much easier to learn then.

src/nimblepkg/version.nim
```diff
@@ -153,6 +153,46 @@ proc makeRange*(version: string, op: string): VersionRange =
...
+    if numbers[n] == "0":
+      inc(zeros)
+    else: break
+  let distanceToMajor = if majorCompatibility: 0 else: 1
+  var excludedPosition = min(numbers.len - 2, zeros + distanceToMajor)
+  if zeros == 2: 
+    excludedPosition = max(2, excludedPosition)
+  numbers[excludedPosition] = $(numbers[excludedPosition].parseInt() + 1)
+  var zeroPosition = excludedPosition + 1
+  while(zeroPosition < numbers.len):
```
dom96, line 191:
```suggestion
  while zeroPosition < numbers.len:
```

dom96 (approved):
Awesome, thank you!

Approving, but I'd like to see some docs for this, can you write some explaining what these new operators mean exactly in the readme? :)

marcomq:
Thx for the review :) I will update the readme. 
After checking the semver documentation again, I have to check if everything is still correct for 2 or 4 digit version numbers. Maybe I also need to add some test cases.

marcomq:
Sry - after creating some documentation, I made some fixes and refactored a lot of code. However, I think the solution is better than my previous one. Let me know if I should make some additional changes.

dom96 (approved):
Looks good, some suggestions for follow up:

* Create tests that verify the version range created for these new intersections, I think that is a much clearer way to test these :)


Might be possible to simplify `getNextIncompatibleVersion`, but let's get it merged, we can iterate later. As long as we have tests it's all good :)

Issue #910, opened 2021-04-17
Nimble Publish Command Security Leak


henryas:
I used `nimble publish` command to publish my first nimble package recently. Nimble automatically added my github security token into the nimble package list and pushed it to github. It was something like this.
```
"url": "https://15b1acb3d21ebde45e8ab271e72fb4a75ad40143:x-oauth-basic@github.com/henryas/nptr"
```
Github automatically detected the leak and revoked the token.

It is better for Nimble to ask for the url, than trying to figure things out on its own and expose someone's security credential. This needs fixing.

Araq:
Strange, Nimble never did that for me... How to reproduce? Just "nimble publish"?

henryas:
Yup. Just `nimble publish`, and answered the questions there. I used the built-in terminal in the vscode. You can check the initial commit there: https://github.com/nim-lang/packages/pull/1870 .

dom96:
How did you get a Git repo with such a URL? Sounds like a pretty bad idea to include a token in the URL, we have SSH keys for a reason.

henryas:
I don't know how such URL came up. I set up Github like ages ago and I don't remember what I did back then. You can tell that from the format of the token that it was Github's old OAuth token. They don't use that format anymore. My guess is that I set it up to work with Github's OAuth, which I believe is still one of Github's recommended authentication methods. So far, I never have to deal with such URLs. It is transparently done behind the scene.

I was learning about Nim through the "Nim in Action" book, which is an excellent book by the way. I used the `nimble publish` command. I answered the prompted questions about the name, tags, license, but it never asked about any URL. Then, nimble automatically pushed a commit to the Github. I did not notice about the leak until one of the reviewers asked about it, and suggested some changes. I checked my email and found Github has revoked the token because it was posted online. So somehow nimble manages to dig out such information from my machine. 

I think it would be nice if Nimble should explicitly ask for the URL because people may connect to Github through a different means (eg. SSH) from the way the general public should download the module.

dom96:
Looking at GitHub's docs, there is a warning not to clone with these URLs which is what I guess you (or another app did): https://docs.github.com/en/github/extending-github/git-automation-with-oauth-tokens#step-2-clone-a-repository

![image](https://user-images.githubusercontent.com/246651/115122812-d9616a80-9fb1-11eb-9a31-27538f31bf25.png)

[...]

PR #911, opened 2021-04-20
possible fix for #910

henryas:
Possible fix for issue https://github.com/nim-lang/nimble/issues/910

### DIFF
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -202,15 +202,15 @@ proc publish*(p: PackageInfo, o: Options) =
       downloadMethod = "git"
     let parsed = parseUri(url)
 
-    # check for any username or password in the URL
-    if parsed.username != "" or parsed.password != "":
-      display("Error: ", "cannot publish the repository URL because it contains username and/or password. Fix the remote URL (Hint: \"git remote -v\").", Error, HighPriority)
-      quit(1)
-
     if parsed.scheme == "":
       # Assuming that we got an ssh write/read URL.
       let sshUrl = parseUri("ssh://" & url)
       url = "https://" & sshUrl.hostname & "/" & sshUrl.port & sshUrl.path
+    elif parsed.username != "" or parsed.password != "":
+      # check for any confidential information
+      raise newException(NimbleError,
+        "Cannot publish the repository URL because it contains username and/or password. Fix the remote URL. Hint: \"git remote -v\"")
+
   elif dirExists(os.getCurrentDir() / ".hg"):
     downloadMethod = "hg"
     # TODO: Retrieve URL from hg.
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -208,6 +208,7 @@ proc publish*(p: PackageInfo, o: Options) =
       url = "https://" & sshUrl.hostname & "/" & sshUrl.port & sshUrl.path
     elif parsed.username != "" or parsed.password != "":
       # check for any confidential information
+      # TODO: Use raiseNimbleError(msg, hintMsg) here
       raise newException(NimbleError,
         "Cannot publish the repository URL because it contains username and/or password. Fix the remote URL. Hint: \"git remote -v\"")
 
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -155,13 +155,6 @@ proc editJson(p: PackageInfo; url, tags, downloadMethod: string) =
   })
   writeFile("packages.json", contents.pretty.cleanupWhitespace)
 
-proc containSensitiveInformation(url: string): bool =
-  ## URL may contain username and password. This proc attempts to detect it.
-  # look for any '@' or '?' character in the url
-  for character in url:
-    if character == '@' or character == '?': return true
-  return false
-
 proc publish*(p: PackageInfo, o: Options) =
   ## Publishes the package p.
   let auth = getGithubAuth(o)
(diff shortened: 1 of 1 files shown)


src/nimblepkg/publish.nim
```diff
@@ -155,6 +155,14 @@ proc editJson(p: PackageInfo; url, tags, downloadMethod: string) =
   })
   writeFile("packages.json", contents.pretty.cleanupWhitespace)
 
+proc containSensitiveInformation(url: string): bool =
+  ## URL may contain username and password. This proc attempts to detect it.
+  # look for any '@' character in the url
+  for character in url:
+    if character == '@':
+      return true
+  return false
```
Araq, line 164:
```suggestion
  '@' in url
```

dom96 (changes requested):
Instead of this just strip the username and password using the `uri` module.

henryas:
So which do you guys prefer? If the url is dirty, ask the user for a new one, or attempt to fix it as @dom96 suggested?

As for @dom96 idea, I was thinking of deleting everything between "https://" up to the first "@" (inclusive), but then I am concerned that it may not always be correct since I do not know the exhaustive list of possible combination in the url.

henryas:
In addition, there is another possibility in which passwords and tokens may be supplied in the url, which is via _http get_. I don't think github allows that and it is not recommended to do that, but some other sites might. I don't think it is possible to properly scrub such information in _http get_ urls. How do you know which argument is the password/token?

I am thinking to just detect for any '@' or '?' in the url, and ask the user if any such characters is found. What do you think?

Araq:
> I am thinking to just detect for any '@' or '?' in the url, and ask the user if any such characters is found. What do you think?

I agree with this solution.

dom96:
In the URI RFCs the data between `https://` and `@` is referred to as the `username` and `password`. You can see the `uri` module in Nim's stdlib also refers to these in the same way: https://nim-lang.org/docs/uri.html#Uri

You don't need to do any parsing yourself, just parseUri the url, clear `username` and `password`, then call `$` on it. 

> In addition, there is another possibility in which passwords and tokens may be supplied in the url, which is via http get. I don't think github allows that and it is not recommended to do that, but some other sites might. I don't think it is possible to properly scrub such information in http get urls. How do you know which argument is the password/token?

This isn't something that any sane website would allow, we shouldn't be detecting for this.

dom96:
Actually, thinking about it some more. I think we should just quit if the URL contains any passwords, this is because it implies that the URL won't work without them, the user should change their git remote.

So I would say:

* Call `parseUri`
* Check for `password`
* if it's not empty: `raise NimbleError("Your git remote contains a password, only public repositories can be published.")`

You can add a hint to the error as well to instruct the user to run `git remote show origin`.

henryas:
I have a couple of questions: what to do when the url contains password and the detection algorithm.

First, if the url contains a password or an access token, based on the feedback so far, we have several options:
1. Terminate `nimble publish` and tell the user to fix their remote url.
2. Proceed with `nimble publish` but ask the user to enter a new url.
3. Proceed with `nimble publish`, ask the user to enter a new url, and drop a hint that their remote url contains security risk. I am not sure if it is our responsibility to do this though. Such url is probably not designed to be read by a third party application such as _nimble_.  Adding a quote from https://www.oauth.com/oauth2-servers/access-tokens/: 

     > Access tokens must be kept confidential in transit and in storage. The only parties that should ever see the access token are the application itself, the authorization server, and resource server. The application should ensure the storage of the access token is not accessible to other applications on the same device. The access token can only be used over an https connection, since passing it over a non-encrypted channel would make it trivial for third parties to intercept.

I personally prefer option no 2.

Second, regarding detection algorithm, wouldn't it be more efficient to detect the `@` character than to parse the whole uri into their individual classifications before deciding whether a password exists? Is there any case where merely detecting '@' will fail?

Regarding _http get_, I did a little research and it turns out it is fairly common to have something like `https://<base url>?access_token=<token>` and Twitter allows it. What I am not clear is whether, in Twitter's case, it is a semi-permanent token (like in github) or a session token. Upon reflection, `http://<username>:<password>@hostname.com/path` is just as risky as `https://<base url>?access_token=<token>`. They both involve encoding sensitive information into the url and the assumption is that the url does not get intercepted. In our case, _nimble_ intercepts the url from a misconfigured git.  

Do we need to cover for the case of _http get_? I am ambivalent about this. I am fine either way whether we cover _http get_ or not.

Let's decide on this and hopefully we can reach a consensus and get this over with.

dom96:
I prefer (1) because like your oauth quote supports: other applications being able to access the oauth token is a security risk. Nimble should scream at the user about it, not silently ignore it and ask for a git URL.

> Second, regarding detection algorithm, wouldn't it be more efficient to detect the @ character than to parse the whole uri into their individual classifications before deciding whether a password exists? Is there any case where merely detecting '@' will fail?

We're not parsing gigabytes of data here. Don't optimise prematurely. Parsing the URL will take care of edge cases that we may not predict and it's much clearer about what's being done.

> They both involve encoding sensitive information into the url and the assumption is that the url does not get intercepted. In our case, nimble intercepts the url from a misconfigured git.

If your URI is accessible in plaintext inside a commonly accessed file then there is no such assumption.

But if you really want you can also make Nimble quit when the `query` field is non-empty. I don't think we have a single package that requires it anyway so it's unlikely to be a problem and it may just save someone from exposing some silly URL.

src/nimblepkg/publish.nim
```diff
@@ -201,10 +201,16 @@ proc publish*(p: PackageInfo, o: Options) =
       if url.endsWith(".git"): url.setLen(url.len - 4)
       downloadMethod = "git"
     let parsed = parseUri(url)
+
     if parsed.scheme == "":
       # Assuming that we got an ssh write/read URL.
       let sshUrl = parseUri("ssh://" & url)
       url = "https://" & sshUrl.hostname & "/" & sshUrl.port & sshUrl.path
+    elif parsed.username != "" or parsed.password != "":
+      # check for any confidential information
```
dom96, line 210:
```suggestion
      # check for any confidential information
      # TODO: Use raiseNimbleError(msg, hintMsg) here
```

PR #912, opened 2021-04-21
tasks: list tasks with proper alignment

kraptor:
This is a small QoL patch.

Currently, `nimble` lists tasks without proper alignment between the name and the description of the task:

```
$ nimble tasks
build_debug        Build debug version
build_release        Build release version
build_profiler        Build with profiler
build_all        Build all versions
clean        Clean all build files
```

This patch adds the ability to collect the max length of the tasks names (with a minimum of 8 characters), then render descriptions properly aligned:

```
$ ./nimble tasks
build_debug     Build debug version
build_release   Build release version
build_profiler  Build with profiler
build_all       Build all versions
clean           Clean all build files
```

I tried to keep the same code style as the code aroudn. Please, let me know if there is anything else you'd like to improve/change.

**NOTE**: this change may *potentially* break scripts dealing with `nimble tasks` output, as it does inject two spaces between task name and task description, instead of a bunch of spaces like before.

The following tests did not pass on master before, not sure if because of my installation of Nim (using 1.5.1-latest) or my system:
* nimscript evaluation error message

### DIFF
--- a/src/nimblepkg/nimscriptapi.nim
+++ b/src/nimblepkg/nimscriptapi.nim
@@ -4,7 +4,7 @@
 ## This module is implicitly imported in NimScript .nimble files.
 
 import system except getCommand, setCommand, switch, `--`
-import strformat, strutils, tables
+import strformat, strutils, tables, sequtils
 export tables
 
 when (NimMajor, NimMinor) < (1, 3):
@@ -32,7 +32,7 @@ var
   foreignDeps*: seq[string] = @[] ## The foreign dependencies. Only
                                   ## exported for 'distros.nim'.
 
-  nimbleTasks: seq[string] = @[]
+  nimbleTasks: seq[tuple[name, description: string]] = @[]
   beforeHooks: seq[string] = @[]
   afterHooks: seq[string] = @[]
   flags: Table[string, seq[string]]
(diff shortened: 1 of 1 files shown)


src/nimblepkg/nimscriptapi.nim
```diff
@@ -178,10 +184,12 @@ template task*(name: untyped; description: string; body: untyped): untyped =
   proc `name Task`*() = body
 
   nimbleTasks.add astToStr(name)
+  if astToStr(name).len > nimbleTasksMaxNameLen:
+    nimbleTasksMaxNameLen = astToStr(name).len
```
dom96, line 188: Nit: you can do this in `onExit` and save the need for another global variable

src/nimblepkg/nimscriptapi.nim
```diff
@@ -178,10 +184,12 @@ template task*(name: untyped; description: string; body: untyped): untyped =
...
+  nimbleTasksDescriptions.add description
```
dom96, line 189: Maybe also turn `nimbleTasks` into a list of tuples to keep the names and descriptions together?

src/nimblepkg/nimscriptapi.nim
```diff
@@ -216,3 +224,6 @@ proc getPkgDir*(): string =
   result = projectFile.rsplit(seps={'/', '\\', ':'}, maxsplit=1)[0]
 
 proc thisDir*(): string = getPkgDir()
+
+
+
```
dom96, line 229:
```suggestion
```

dom96 (changes requested):

src/nimblepkg/nimscriptapi.nim
kraptor: Done.

src/nimblepkg/nimscriptapi.nim
kraptor: Done.

src/nimblepkg/nimscriptapi.nim
kraptor: Done. I wanted to avoid an extra loop, but normally there are few tasks, so not critical.

kraptor:
Ok, I see what is failing... let me fix printSeqIfLen template...

kraptor:
I squashed the changes for easy review @dom96, everything seems ok now.

PR #920, opened 2021-05-23
publish: Ensure pulls happen to the correct branch

kaushalmodi:
If a user has set the default branch to anything other than "master",
'nimble publish' fails.

This commit fixes that by ensuring that the git pull followed by git
init happens to the default branch used by upstream https://github.com/nim-lang/packages.

For example, if the user has this in their `~/.gitconfig`:

```
[init]
	defaultBranch = main
```

`nimble publish` fails with this error:

```
error: src refspec master does not match any
error: failed to push some refs to 'https://github.com/kaushalmodi/packages'
```

I have confirmed that after this commit, that error gets resolved. The critical change in this commit fixing this issue if ensuring that we first check out the default branch in user's git repo if the default branch != repo's default branch ("master").

### DIFF
--- a/src/nimblepkg/publish.nim
+++ b/src/nimblepkg/publish.nim
@@ -20,6 +20,7 @@ const
   ApiKeyFile = "github_api_token"
   ApiTokenEnvironmentVariable = "NIMBLE_GITHUB_API_TOKEN"
   ReposUrl = "https://api.github.com/repos/"
+  defaultBranch = "master" # Default branch on https://github.com/nim-lang/packages
 
 proc userAborted() =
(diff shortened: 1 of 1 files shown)


kaushalmodi:
Note that a better way to do this would be to do `git init --initial-branch="master"` but that would fail if the is using git older than 2.28.0.

- [git init docs in git v2.28.0](https://git-scm.com/docs/git-init/2.28.0)
- [git init docs in git v2.27.0](https://git-scm.com/docs/git-init/2.27.0) (`--initial-branch` option missing in this version)

dom96 (approved):

PR #934, opened 2021-08-22
change `nimbleDir` to xdg specification

dm9pZCAq:
**note**: in `devel` branch there are [getCacheDir](https://nim-lang.github.io/Nim/os.html#getCacheDir%2Cstring) `proc`, but in `1.4.8` there is not

and at next `nim` release we can just do:
```nim
result.nimbleDir = getCacheDir("nimble")
```

### DIFF
--- a/src/nimblepkg/config.nim
+++ b/src/nimblepkg/config.nim
@@ -18,7 +18,14 @@ type
     path*: string
 
 proc initConfig(): Config =
-  result.nimbleDir = getHomeDir() / ".nimble"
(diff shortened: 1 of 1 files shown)


dom96 (changes requested):
The default location for the NIMBLE_DIR is ~/.nimble by design. You often need access to this directory so I don't think changing it is the right approach here. If you wish to move it in your system feel free to set the `NIMBLE_DIR` env var.

PR #943, opened 2021-09-20
Update outdated documentation

bobeff:
The part of the documentation about the handling of transitive package dependencies was no longer correct since commit dfeed75. Now it is updated to describe the new behavior.


haxscramper:
Doc update must also include explanation for nimble develop from https://forum.nim-lang.org/t/8404

PR #946, opened 2021-09-28
Change the name of the "pkgs" directory to "pkgs2"

bobeff:
The names of the "nimbledata.json" file and "pkgs" directory are changed correspondingly to "nimbledata2.json" and "pkgs2" in order to avoid the need of removing the Nimble cache when changing between the new and old Nimble versions.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -816,9 +816,9 @@ Library packages are likely the most popular form of Nimble packages. They are
 meant to be used by other library or binary packages.
 
 When Nimble installs a library, it will copy all of its files
-into ``$nimbleDir/pkgs/pkgname-ver``. It's up to the package creator to make sure
-that the package directory layout is correct, this is so that users of the
-package can correctly import the package.
+into ``$nimbleDir/pkgs2/pkgname-ver-checksum``. It's up to the package creator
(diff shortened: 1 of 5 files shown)


Araq:
Thank you!

pietroppeter:
I think this change (pkgs -> pkgs2) should be reflected not only by updating the documentation to be correct, but also should be highlighted as a major changing happening in 0.14 (there probably should be a section for changes introduced in 0.14). at the very least it should be mentioned in the changelog.

note also that this PR contains another commit (was it intended to end up here?) where description becomes an optional field instead of required. documentation and changelog should also be updated for this! (see also recent issue https://github.com/nim-lang/nimble/issues/947)

pietroppeter:
had a discussion with @Araq on Discord #nimble channel, the rationale for approving making description field optional is:

> the problem with "description" is that it's redundant with a readme.md file which is pretty much always there thanks to github
> and in the past there were discussions of many things nimble thinks it needs but which are redundant, version info (require git tags), author (part of the git metadata), description (readme.md)
> Nimble should be "smart" and simply use what is available and not enforce code duplication. (readme.md is usually up to date and the Nimble description isn't)

PR #952, opened 2021-10-11
Fix getting download dir name for develop

bobeff:
When getting the download directory name for a package that we want to put in develop mode, a trailing slash in the URL or the subdir leads to getting an empty string. To fix the problem we remove the trailing slash from them at the start of the `getDevelopDownloadDir` procedure.


dom96 (approved):
Makes sense. But can we get some quick `assert` tests to test `getDevelopDownloadDir`?

bobeff:
> 
> 
> Makes sense. But can we get some quick `assert` tests to test `getDevelopDownloadDir`?

@dom96 Done.

PR #962, opened 2021-11-30
Fix #953

### LINKED ISSUE #953: `nimble install` doesn't respect newly downloaded `packages.json`
When a package is not found,
and prompted to download a new list from the internet,
the package will still be reported as "not found'
(even if it was in the newly downloaded list).

I experienced this with `chipmunk7`,
which was recently added to the list.

`nimble install chipmunk7` worked the second time I tried the command.

ire4ever1190:
Fixes [issue](https://github.com/nim-lang/nimble/issues/953) where nimble didn't search refreshed package list when trying to install a package

Current behavior
 
image

Behavior after PR

image

Is the way I did it in the PR good or should I instead implement a proc which clears `gPackageJson` table?

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -603,14 +603,14 @@ proc processLockedDependencies(pkgInfo: PackageInfo, options: Options):
       result.incl installDependency(pkgInfo, downloadResult, options)
 
 proc getDownloadInfo*(pv: PkgTuple, options: Options,
-                      doPrompt: bool, clean = false): (DownloadMethod, string,
+                      doPrompt: bool, ignorePackageCache = false): (DownloadMethod, string,
                                         Table[string, string]) =
   if pv.name.isURL:
     let (url, metadata) = getUrlData(pv.name)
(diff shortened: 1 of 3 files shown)


xflywind:
I have seen many Nim users cannot use nimble due to this error. I suffer from this error too.

src/nimble.nim
```diff
@@ -603,14 +603,14 @@ proc processLockedDependencies(pkgInfo: PackageInfo, options: Options):
       result.incl installDependency(pkgInfo, downloadResult, options)
 
 proc getDownloadInfo*(pv: PkgTuple, options: Options,
-                      doPrompt: bool): (DownloadMethod, string,
+                      doPrompt: bool, clean = false): (DownloadMethod, string,
```
dom96, line 606:
```suggestion
                      doPrompt: bool, ignorePackageCache = false): (DownloadMethod, string,
```

dom96 (changes requested):
Thanks! Just one request for changes: let's have clearer identifier names :)

Issue #261, opened 2016-10-11
Feature request: Dry run mode


mcclure:
Many package managers have a `--dry-run` mode, where nothing is installed and instead only prints out the paths it WOULD have installed things to, had it installed.

I don't know if this feature is compatible with nimble's model, but it *is* a nice feature that would improve nimble if it were present. I just found myself hovering over the manual install commands for nimble on a new machine but hesitated, thinking, wait, what does this actually *do*? It would have been nice to run `install --dry-run` and verify it only installs to paths that won't make a mess.

dom96:
Sure, sounds like a good feature.

PR #964, opened 2021-12-10
Add --dryRun option

CyberTailor:
I'm not able to test this because nimble fails to build (`tools.nim(11, 14) Error: cannot open file: compiler/nimblecmd`) even with Nim compiled from devel branch, so testing is required.

This PR closes issue #261

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -145,8 +145,9 @@ proc processFreeDependencies(pkgInfo: PackageInfo, options: Options):
   # them added if the above errorenous condition occurs
   # (unsatisfiable dependendencies).
   # N.B. NimbleData is saved in installFromDir.
-  for i in reverseDependencies:
-    addRevDep(options.nimbleData, i, pkgInfo)
+  if not options.dryRun:
+    for i in reverseDependencies:
+      addRevDep(options.nimbleData, i, pkgInfo)
 
 proc buildFromDir(pkgInfo: PackageInfo, paths: HashSet[string],
                   args: seq[string], options: Options) =
@@ -196,7 +197,7 @@ proc buildFromDir(pkgInfo: PackageInfo, paths: HashSet[string],
             [pkginfo.basicInfo.name, bin, pkgInfo.backend], priority = HighPriority)
 
     let outputDir = pkgInfo.getOutputDir("")
-    if not dirExists(outputDir):
+    if not dirExists(outputDir) and not options.dryRun:
       createDir(outputDir)
 
     let input = realDir / src.changeFileExt("nim")
@@ -206,7 +207,8 @@ proc buildFromDir(pkgInfo: PackageInfo, paths: HashSet[string],
       getNimBin(options).quoteShell, pkgInfo.backend, join(args, " "),
       outputOpt, input.quoteShell]
     try:
-      doCmd(cmd)
+      if not options.dryRun:
+        doCmd(cmd)
       binariesBuilt.inc()
     except CatchableError as error:
       raise buildFailed(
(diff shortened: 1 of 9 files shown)


tests/tuninstall.nim
```diff
@@ -42,6 +47,45 @@ suite "uninstall":
...
+    check execNimbleYes("--dryRun", "uninstall", "packagea").exitCode == QuitSuccess
+    check execNimbleYes("--dryRun", "uninstall", "PackageA").exitCode == QuitSuccess
+
+    check execNimbleYes("--dryRun", "uninstall", "PackageB").exitCode == QuitSuccess
+
+    check execNimbleYes("--dryRun", "uninstall", "PackageA@0.2", "issue27b").exitCode !=
+        QuitSuccess
+
+    check execNimbleYes("--dryRun", "uninstall", "nimscript").exitCode == QuitSuccess
+
```
dom96, line 88: Does this test to verify that a package wasn't uninstalled due to the `--dryRun`? Maybe I'm missing it but I don't see it. Please call out the part of this test that tests for this with a comment in the code.

src/nimblepkg/nimscriptexecutor.nim
```diff
@@ -38,7 +38,8 @@ proc execCustom*(nimbleFile: string, options: Options,
     raise nimbleError("Pre-hook prevented further execution.")
 
   if not nimbleFile.isNimScript(options):
-    writeHelp()
+    writeHelp(false)
+    raise nimbleQuit(QuitFailure)
```
dom96, line 42: Doesn't `writeHelp` already raise?

src/nimble.nim
```diff
@@ -1738,7 +1765,8 @@ when fileExists("{nimblePathsFileName}"):
     writeFile = true
 
   if writeFile:
-    writeFile(nimbleConfigFileName, fileContent)
+    if not options.dryRun:
+      writeFile(nimbleConfigFileName, fileContent)
     displayInfo(&"\"{nimbleConfigFileName}\" is set up.")
```
dom96, line 1770: seeing this with `--dryRun` would be quite confusing

src/nimble.nim
```diff
@@ -1263,7 +1282,8 @@ proc updatePathsFile(pkgInfo: PackageInfo, options: Options) =
   for path in paths:
     pathsFileContent &= &"--path:{path.escape}\n"
   var action = if fileExists(nimblePathsFileName): "updated" else: "generated"
-  writeFile(nimblePathsFileName, pathsFileContent)
+  if not options.dryRun:
+    writeFile(nimblePathsFileName, pathsFileContent)
   displayInfo(&"\"{nimblePathsFileName}\" is {action}.")
```
dom96, line 1287: this will seem confusing with --dryRun

dom96 (changes requested):
This looks good, but I fear it will be quite fragile. It's unclear which procs handle `dryRun` and which don't. For example you have `setupNimbleConfig` handling it, but then below where it's called you see another check for `dryRun`. This means we might end up with `dryRun` not being correctly handled in the future because someone not familiar with the code forgets to use it, or uses it more than necessary.

It's definitely non-trivial to make this safer though unfortunately. The only thing I can think of is to create Nimble-specific wrappers for all procs that write/remove/etc files and execute commands. Tbh maybe we should do this? We already have `doCmd`.

My other concern is that we will see messages the imply actions were taken, even with the `dryRun`. We need to make it clear that they weren't taken.

CyberTailor:
Closing as this implementation is too fragile

Issue #956, opened 2021-11-07
`nimble install`: support staged installation


CyberTailor:
Package managers are usually sandboxed (they don't allow direct installation to system root but install to some intermediate directory instead).

This approach works for most common build systems. They

* Have "configure" to find dependencies in the system root
* Install to [$(DESTDIR)](https://www.gnu.org/software/automake/manual/html_node/DESTDIR.html)

But Nimble checks dependencies and installs files in the same command, so I have to sed(1) out all "requires" lines to make it work with Portage.

Some Nimble option can be introduced for manual dependency resolution.

CyberTailor:
Moreover, `nimble install` builds the package even when `nimble build` has been already run, so there's no way to separate compilation and installation. If it was possible, I could sed(1) out all "requires" lines in the \*.nimble file before installation.

PR #965, opened 2021-12-10
nimble install: add --destDir option

CyberTailor:
Requires testing to make sure it works as intended.

This PR closes issue #956

### DIFF
--- a/src/nimblepkg/options.nim
+++ b/src/nimblepkg/options.nim
@@ -70,6 +70,7 @@ type
       devActions*: seq[DevelopAction]
       path*: string
       noRebuild*: bool
+      destDir*: string
       withDependencies*: bool
         ## Whether to put in develop mode also the dependencies of the packages
         ## listed in the develop command.
@@ -102,6 +103,7 @@ Commands:
                [-d, --depsOnly]   Install only dependencies.
                [-p, --passNim]    Forward specified flag to compiler.
                [--noRebuild]      Don't rebuild binaries if they're up-to-date.
+               [--destDir:path]   Prepend path to each installed target file.
   develop      [pkgname, ...]     Clones a list of packages for development.
                                   Adds them to a develop file if specified or
                                   to `nimble.develop` if not specified and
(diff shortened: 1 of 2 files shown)


dom96 (changes requested):
Since we already have `--nimbleDir`, why do we need this?

CyberTailor:
> Since we already have `--nimbleDir`, why do we need this?

It's for package managers, which are usually sandboxed and don't allow direct installation.

`--nimbleDir` (e.g. /opt/nimble) is used to find dependencies in the system root and `--destDir` (e.g. /var/tmp/portage/dev-nim/example/image) is used to install package files to some intermediate location.

If I just set `--nimbleDir` to /var/tmp/..., nimble will be unable to find dependencies.

dom96:
I'm not sure what you mean. If you set `nimbleDir` to `/var/tmp/nimble` it will install all packages in there and find deps just fine

dom96:
I can appreciate that. But it doesn't answer my question: why can't you simply use `--nimbleDir`?

CyberTailor:
> why can't you simply use `--nimbleDir`?

Because there are no dependencies in `/var/tmp/portage/dev-nim/example/image/opt/nimble`, they are in `/opt/nimble`. Things are only installed into this temporary directory, it's not the actual system root.

If I set `--nimbleDir`, nimble will try to find dependencies in `/var/tmp/...` and fail.  
If I don't set it, nimble will attempt to write directly to `/opt/nimble` and fail because system package manager's sandbox won't allow such operation.

Flowdalic:
I think @CyberTailor already did a good job explaining the motivation. But I'd like to add that `DESTDIR` is long standing concept for staged installs. It is defined GNU coding standards - see https://www.gnu.org/prep/standards/html_node/DESTDIR.html#DESTDIR for a good summary [1] - but is found in various other build systems.

<sup>1: Note that it concept probably does predate GNU coding standards</sup>

dom96:
Isn't the real problem here that Nimble doesn't actually have support for multiple Nimble dirs? What we really want is to add support for this first I think. What follows then is a way to specify which Nimble dir packages should be installed in, not this `destDir` which while it seems like it follows GNU conventions, feels like more of a hack to me :)

CyberTailor:
Got it. I'm going to implement support for multiple Nimble dirs then.

PR #966, opened 2021-12-11
Add --noRebuild option for install

CyberTailor:
Needs testing to make sure it works as expected.

This PR allows to separate build and installation, this is needed for installing nim packages using source-based system package managers like portage.


dom96 (approved):
Epic. Thanks!

PR #969, opened 2021-12-17
Support multiple Nimble dirs

CyberTailor:
This PR closes issue #956. It is more complex than PR #965 (`--destDir`) but hypothetically solves wider range of use cases.

**Important:** only the main Nimble dir can be modified, the others are read-only (this is necessary for packaging in Linux distributions).

Some regressions:
* `uninstall --inclDeps` can no longer be used with `--nimbleDir` - use `--excludeNimbleDir` to leave only one Nimble dir (workaround applied in test suite)
* `assert result.metaData.vcsRevision == notSetSha1Hash` removed because it no longer works for some reason

Some edge cases are probably not covered.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -57,7 +57,7 @@ proc refresh(options: Options) =
 
 proc initPkgList(pkgInfo: PackageInfo, options: Options): seq[PackageInfo] =
   let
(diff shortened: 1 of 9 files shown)


CyberTailor:
Should I remove `--excludeNimbleDir` option and do the following instead: add `options.config.nimbleDir` only if there are no other Nimble dirs?

Issue #957, opened 2021-11-09
`nimble sync` reports checksum mismatch on Windows


zer0pwned:
Hi, I was trying to build a nim build system with Github Action and I noticed that the new nimble support lock/sync feature which is quite useful. 

The nimble (v0.14.0) sync works fine on Linux and MacOS. But it reports error on Windows build system. See the reported error below. You can find the raw Github Action log over here https://github.com/zer0pwned/nim-pipeline-system/runs/4147410444?check_suite_focus=true. 

I tried to use the same nim and nimble version (1.6.0 and 0.14.0) on my local Windows machine and I got the same error message. Is there anything wrong with my setup or it's just the nimble checksum mechanism is broken on Windows?

```
/c/Users/runneradmin/.nimble/pkgs/nimble-#0a23c44c/nimble
Nim Compiler Version 1.6.0 [Windows: amd64]
Compiled at 2021-10-19
Copyright (c) 2006-2021 by Andreas Rumpf

active boot switches: -d:release
nimble v0.14.0 compiled at 2021-11-09 03:02:14
git hash: 0a23c44cd8d6c7856d5da155a32298bb1a6ca8e9

Downloading https://github.com/status-im/nim-stew using git
       Tip: 4 messages have been suppressed, use --verbose to show them.
nimble.nim(553)          downloadDependency

    Error:  Downloaded package checksum does not correspond to that in the lock file:
        ...   Package:           stew@v.0.1.0@r.b002c1fad025c6b5fe1f6af1fbb7502dc3173beb
        ...   Checksum:          9cc42eb82a1769bf22adeff7704e109a6e471a5b
        ...   Expected checksum: b81ecd17eb85e17ad38a0d173484dfba371ff940
Error: Process completed with exit code 1.
```

zer0pwned:
Ok I think I figured out what's going on. By default, git on Windows configs the autocrlf option to be true, including Github action Windows runner. Nimble uses git to checkout the dependencies and thus suffer this checksum mismatch if autocrlf option is true. To get around it user can manually setup this option to be input or even false but I do think nimble probably should handle this otherwise it will be kind of annoying to regular developers. 

```
user@builder MINGW64 /c/runners/actions/_work/builder/builder (develop)
$ git config --global core.autocrlf true

user@builder MINGW64 /c/runners/actions/_work/builder/builder (develop)
$ ~/.nimble/pkgs/nimble-#head/nimble -l sync
   Warning: Using project local deps mode
[...]

PR #974, opened 2021-12-22
Fix the `doClone` procedure to clone a repository as is

bobeff:
To have the right package checksum when cloning a Git repository, we have to clone it without changing the line endings according to the operating system.

Fixes nim-lang/nimble#957


dom96 (approved):

bobeff:
This fix seems to broke the tests exactly in the way of the bug it was intended to fix. :D

PR #977, opened 2022-01-09
Enable unit tests for GitHub Actions

CyberTailor:


dom96 (approved):
lgtm

bobeff:
@dom96 @CyberTailor Those unit tests are already enabled when the `tester.nim` module is being run. It imports `tmoduletests.nim` which runs the unit tests, so this change is redundant.

dom96:
oh cool. We should revert this then, while we're at it I'd say we should modify `tmoduletests` to compile and run each module in Nimble's sources, otherwise we'll end up forgetting to add new modules in there. What do you think?

PR #983, opened 2022-02-03
Fix create directory behavior

locriacyber:
Before this fix, `nimble init` will create a new directory if current directory name is `a-b`, because the inferred package name is `a_b`.

This fixed the behavior to align with what was in #503, so `nimble init` will never create a new directory.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -963,6 +963,7 @@ proc init(options: Options) =
     raise nimbleError("Please install git or mercurial first")
 
   # Determine the package name.
+  let hasprojectname = options.action.projName != ""
   let pkgName =
     if options.action.projName != "":
       options.action.projName
@@ -974,7 +975,7 @@ proc init(options: Options) =
 
   # Determine the package root.
   let pkgRoot =
-    if pkgName == os.getCurrentDir().splitPath.tail:
+    if not hasprojectname:
       os.getCurrentDir()
     else:
       os.getCurrentDir() / pkgName
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -975,7 +975,7 @@ proc init(options: Options) =
 
   # Determine the package root.
   let pkgRoot =
-    if not hasprojectname:
+    if not hasProjectName:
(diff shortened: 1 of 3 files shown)


src/nimble.nim
```diff
@@ -963,6 +963,7 @@ proc init(options: Options) =
     raise nimbleError("Please install git or mercurial first")
 
   # Determine the package name.
+  let hasprojectname = options.action.projName != ""
```
dom96, line 966:
```suggestion
  let hasProjectName = options.action.projName != ""
```

dom96 (changes requested):
I think this needs some tests, any chance you could add some? :)

locriacyber:
I don't know how to add tests to nimble. It would be something like this:
```
mkdir a-b
cd a-b
nimble init # anwser all
test -e a_b.nimble # test file exist in pwd
```

dom96:
Lots of examples in here https://github.com/nim-lang/nimble/tree/master/tests

tests/tinitcommand.nim
```diff
@@ -0,0 +1,17 @@
...
+import testscommon
+
+from nimblepkg/common import cd
+
+suite "init":
+  ## https://github.com/nim-lang/nimble/pull/983
+  test "init within directory that is invalid package name will not create new directory":
+    cleanDir(installDir)
+    let tempdir = installDir / "a-b"
+    createDir(tempdir)
```
dom96, line 11: I think you want to use getTempDir here.

dom96 (approved):
Thanks for the test! Once the CI passes this is good to go.

tests/tinitcommand.nim
locriacyber: Fixed. How do I pass CI tests?

dom96: Looks like you've got warnings in your changes: https://github.com/nim-lang/nimble/runs/5265139850?check_suite_focus=true#step:6:221

locriacyber:
@dom96 not merged yet?

PR #1009, opened 2022-07-15
Can sync ignoring deps not present in lock file even if they are in develop file

yyoncho:


dom96 (commented):
Changes seem good, but can you elaborate the situation this covers?

yyoncho:
> Changes seem good, but can you elaborate the situation this covers?

In https://github.com/status-im/nim-workspace we are going to have one develop file that is shared across multiple projects which means that you may have projects in develop mode that are not actually dependencies of the current project. This will result in failure when doing `nimble sync`.

PR #1011, opened 2022-07-25
Allow overriding lock file

yyoncho:


src/nimblepkg/options.nim
```diff
@@ -752,3 +756,10 @@ proc getCompilationBinary*(options: Options, pkgInfo: PackageInfo): Option[strin
 
 proc isInstallingTopLevel*(options: Options, dir: string): bool =
   return options.startDir == dir
+
+proc lockFile*(options: Options, dir: string): string =
+  let lockFile = if options.lockFileName == default(string):
+    defaultLockFileName
+  else:
+    options.lockFileName
+  result = dir / lockFile
```
zah, line 765: This doesn't seem to be handling absolute paths properly.

yyoncho: lockFile is supposed to be relative. If you want I can add a check if the lock file path is absolute.

yyoncho: updated the PR to handle that case.

PR #1017, opened 2022-08-08
Allow locking nim version in nimble.lock

### LINKED ISSUE #953: `nimble install` doesn't respect newly downloaded `packages.json`

yyoncho:
- Fixes #953

Allow having nim as locked dependency.

- I will add unit tests once we agree on the approach and once nimble related
changes in nim are merged (I will link the PR in comment). Ditto for the
documentation.

Here it is the flow:

``` bash
nimble develop nim
nimble lock
```

After that `nimble install` and `nimble build` commands will use the locked
`nim` version

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -332,8 +332,9 @@ proc processAllDependencies(pkgInfo: PackageInfo, options: Options):
   if pkgInfo.lockedDeps.len > 0:
     result = pkgInfo.processLockedDependencies(options)
   else:
-    var requires = pkgInfo.requires & pkgInfo.taskRequires.getOrDefault(options.task)
-    pkgInfo.processFreeDependencies(requires, options, result)
+    pkgInfo.processFreeDependencies(pkgInfo.requires, options, result)
+    if options.task in pkgInfo.taskRequires:
+      pkgInfo.processFreeDependencies(pkgInfo.taskRequires[options.task], options, result)
 
 proc allDependencies(pkgInfo: PackageInfo, options: Options): HashSet[PackageInfo] =
   ## Returns all dependencies for a package (Including tasks)
@@ -604,7 +605,7 @@ proc installDependency(pkgInfo: PackageInfo, downloadInfo: DownloadInfo,
     downloadInfo.vcsRevision)
 
   downloadInfo.downloadDir.removeDir
-  let deps = pkgInfo.lockedDeps[""]
+  let deps = pkgInfo.lockedDeps[noTask]
   for depDepName in downloadInfo.dependency.dependencies:
     let depDep = deps[depDepName]
     let revDep = (name: depDepName, version: depDep.version,
@@ -623,18 +624,16 @@ proc processLockedDependencies(pkgInfo: PackageInfo, options: Options):
 
   let developModeDeps = getDevelopDependencies(pkgInfo, options)
 
-  for task, deps in pkgInfo.lockedDeps:
-    if task in ["", options.task]:
-      for name, dep in deps:
-        if developModeDeps.hasKey(name):
-          result.incl developModeDeps[name][]
-        elif isInstalled(name, dep, options):
-          result.incl getDependency(name, dep, options)
-        elif not options.offline:
-          let downloadResult = downloadDependency(name, dep, options)
-          result.incl installDependency(pkgInfo, downloadResult, options)
-        else:
-          raise nimbleError("Unsatisfied dependency: " & pkgInfo.basicInfo.name)
+  for name, dep in pkgInfo.lockedDepsFor(options):
+    if developModeDeps.hasKey(name):
+      result.incl developModeDeps[name][]
+    elif isInstalled(name, dep, options):
+      result.incl getDependency(name, dep, options)
+    elif not options.offline:
+      let downloadResult = downloadDependency(name, dep, options)
+      result.incl installDependency(pkgInfo, downloadResult, options)
+    else:
+      raise nimbleError("Unsatisfied dependency: " & pkgInfo.basicInfo.name)
 
 proc getDownloadInfo*(pv: PkgTuple, options: Options,
                       doPrompt: bool, ignorePackageCache = false): (DownloadMethod, string,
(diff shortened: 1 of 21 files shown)


src/nimblepkg/packageparser.nim
```diff
@@ -116,43 +116,45 @@ proc validatePackageStructure(pkgInfo: PackageInfo, options: Options) =
...
-                "This will be an error in the future.") %
-              [file & ext, pkgInfo.basicInfo.name, correctDir, dir]
-          hint = ("If '$1' contains source files for building '$2', rename it " &
-                  "to '$3'. Otherwise, prevent its installation " &
-                  "by adding `skipDirs = @[\"$1\"]` to the .nimble file.") %
-              [dir, pkgInfo.basicInfo.name, correctDir]
-        raise validationError(msg, true, hint, true)
+    # nim is used for building the project, thus no need to validate its structure.
+    if not pkgInfo.basicInfo.name.isNim:
+      if dir.len == 0:
```
dom96, line 121: Nit: instead of adding another nesting to this if, please just put the condition into one of the existing if statements.

src/nimblepkg/packageinfo.nim
```diff
@@ -530,6 +528,9 @@ proc hash*(x: PackageInfo): Hash =
 proc getNameAndVersion*(pkgInfo: PackageInfo): string =
   &"{pkgInfo.basicInfo.name}@{pkgInfo.basicInfo.version}"
 
+proc isNim*(name: string): bool =
+  result = name == "nim" or name == "nimrod"
```
dom96, line 532: Nim hasn't been known by that name in long enough to exclude it IMO

src/nimblepkg/tools.nim
```diff
@@ -62,13 +62,8 @@ proc tryDoCmdEx*(cmd: string): string {.discardable.} =
...
   return output
 
-proc getNimBin*: string =
-  result = "nim"
-  if findExe("nim") != "": result = findExe("nim")
-  elif findExe("nimrod") != "": result = findExe("nimrod")
-
 proc getNimrodVersion*(options: Options): Version =
-  let vOutput = doCmdEx(getNimBin(options).quoteShell & " -v").output
+  let vOutput = doCmdEx(getNimBin(options) & " -v").output
```
dom96, line 66: if you're gonna do that then rename `getNimBin` to `getNimBinQuoted`

src/nimble.nim
```diff
@@ -66,7 +66,7 @@ proc initPkgList(pkgInfo: PackageInfo, options: Options): seq[PackageInfo] =
 proc install(packages: seq[PkgTuple], options: Options,
              doPrompt, first, fromLockFile: bool): PackageDependenciesInfo
 
-proc processFreeDependencies(pkgInfo: PackageInfo, options: Options):
+proc processFreeDependencies(pkgInfo: PackageInfo, options: Options, nimAsDepedency = false):
```
dom96, line 69:
```suggestion
proc processFreeDependencies(pkgInfo: PackageInfo, options: Options, nimAsDependency = false):
```

src/nimblepkg/tools.nim
dom96: alternatively just revert this change, quoteShell is much better closest to doCmdExe to ensure we use it (plus other programs like choosenim might be using getNimBin and won't be expecting quoteShell)

dom96 (commented):
Seems fair. Though can you elaborate what will end up in the lock file when `nimble lock` is called?

src/nimblepkg/packageparser.nim
yyoncho: In this case, I will have to add one if in the else clause as well. It will be:

```
if not foo and bar:
  ...
elif not foo
  ...
```
To me this looks less "clean. Let me know what you think.

src/nimblepkg/packageinfo.nim
yyoncho: Ok, will do. I will rename a few functions using nimrod in their name as well.

src/nimblepkg/tools.nim
yyoncho: I will revert this refactoring and keep the `quoteShell` in the client.

src/nimblepkg/packageparser.nim
dom96: ahh, in that case if you wouldn't mind refactoring it out into its own proc that would be the best of both worlds :)

yyoncho: Can you elaborate?

yyoncho:
> Seems fair. Though can you elaborate what will end up in the lock file when `nimble lock` is called?

Here is what it looks like(I have patched packages.json locally to make that work until we have an installable version in main repo)
``` 
  {
  "version": 1,
  "packages": {
    "nim": {
      "version": "1.7.1",
      "vcsRevision": "afc3c24ff713453aea869df0df93e591e9947f24",
      "url": "https://github.com/yyoncho/Nim",
      "downloadMethod": "git",
      "dependencies": [],
      "checksums": {
        "sha1": "33781958ab4835a413519059a03217ee81e56dc3"
      }
    }
  }
}
```

yyoncho:
@dom96 please disregard this PR - refer to https://github.com/nim-lang/Nim/pull/20179#issuecomment-1209010294 

I have to revisit the implementation.

src/nimblepkg/packageparser.nim
dom96: Move the whole `if dir.len == 0:` if statement (its body and the `else`) into a `validatePackageStructure` proc, then only call it under the `if not pkgInfo.basicInfo.name.isNim:`

dom96:
Suggested a solution :)

src/nimblepkg/packageparser.nim
yyoncho: Ok, will do.

yyoncho:
@dom96 applied your comments(hopefully I haven't missed anything).

src/nimblepkg/packageinfo.nim
yyoncho: > Nim hasn't been known by that name in long enough to exclude it IMO

I reverted that part - nimrod is used in the tests so I don't think that this PR is the place to go and fix them.

dom96:
We need to wait for @Araq to give his approval on the Nim PR. So I would say this PR is blocked on that.

arnetheduck:
> Here it is the flow:

why is `nimble develop` needed? `nim` should simply always be part of the lock file, just like any other dep?

yyoncho:
> > Here it is the flow:
> 
> why is `nimble develop` needed? `nim` should simply always be part of the lock file, just like any other dep?

I don't have a strong preference for that. The idea was to avoid managing nim by accident and having it more explicit. Zah suggestion was to do that with flag, but I did it with `nimble develop` which will help fixing the version as well.

yyoncho:
As per my discussion with @arnetheduck I was trying to add flag to make locking possible with `nimble --lock-nim lock` but that will download the latest tagged version which does not have the proper package structure. 

@arnetheduck I think that this feature should be held on till we tag nim version, WDYT?

arnetheduck:
> @arnetheduck I think that this feature should be held on till we tag nim version, WDYT?

I believe this is the case for any nimble dependency, ie nimble will happily download any random version of anything, depending on what is in the repo at the time and what the user happens to have in their global cache. 

I think the two issues can be addressed separately - in particular, I  imagine that if an application wants a specific nim version, it would indicate so in its `nimble` file - until nimble has a proper dep resolver and a `nimble update package` command that allows updating a particular dependency to a particular version, users can set `nim#somehash` as their version requirement and it should hopefully force nimble to do the right thing (or die trying)

Araq:
Fails with:


  Verifying dependencies for nim@1.9.1
/home/runner/work/nimble/nimble/src/nimble.nim(2158) nimble
/home/runner/work/nimble/nimble/src/nimble.nim(2111) doAction
/home/runner/work/nimble/nimble/src/nimble.nim(1695) lock
/home/runner/work/nimble/nimble/src/nimble.nim(1528) validateDevelopDependenciesVersionRanges

    Error:  Some of the develop mode dependencies are with versions which are not in the required by other package's Nimble file range.

yyoncho:
I will take a look, it doesn't fail locally on my side.

yyoncho:
> Fails with:
> 
> Verifying dependencies for nim@1.9.1 /home/runner/work/nimble/nimble/src/nimble.nim(2158) nimble /home/runner/work/nimble/nimble/src/nimble.nim(2111) doAction /home/runner/work/nimble/nimble/src/nimble.nim(1695) lock /home/runner/work/nimble/nimble/src/nimble.nim(1528) validateDevelopDependenciesVersionRanges
> 
> ```
> Error:  Some of the develop mode dependencies are with versions which are not in the required by other package's Nimble file range.
> ```

@Araq It is good now.

PR #1020, opened 2022-08-14
tweak docs to clarify install depsOnly option

elcritch:
So I never realized you could leave out the pkgname and it'd use the local nimble file. It's clear in hindsight but reading the docs and help I always assumed `install` always required the package name. Hopefully this tweak might save others from missing this too.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -180,6 +180,12 @@ Example:
     ...
     nake installed successfully
 
+Packages needed for developing or testing a project can be installed by
+passing the `--depsOnly` option and leaving out the package name.  
+Example:
+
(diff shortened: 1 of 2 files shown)


dom96 (commented):
Nice, thanks. I think it would be good to have this after the paragraph starting "If you don't specify a parameter..." can you move it there?

elcritch:
Certainly, I'll update it.

elcritch:
I started editing but wasn't quite happy with it. Adding the subsection seemed to make it a bit more clearer to the different usages of `install` (package manager vs local package dev).

dom96 (approved):
Thanks!

PR #1022, opened 2022-08-15
Add option to install from nimble file

elcritch:
This PR adds an option to the `install` command to work from a local Nimble file. This makes it just a bit easier to install a package or dependencies in a CI or environment setup scripts. Doing a `cd pkgdir && nimble ...` can be annoying in some CI scripts. I figured an explicit option produced cleaner code and semantics than just checking if a pkgname ended with a ".nimble" extension. Though doing a `local://` or `file://` might another option. 

Really it's not a *necessary* feature and partly I was motived partly by adding a small feature to Nimble to learn about it. If it's handy, great, if not no worries! :) 

Examples:

```sh
./nimble install --depsOnly --nimble-file testpkg/testpkg.nimble 
./nimble install --nimble-file testpkg/testpkg.nimble 
```

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -622,9 +622,13 @@ proc processLockedDependencies(pkgInfo: PackageInfo, options: Options):
     else:
       raise nimbleError("Unsatisfied dependency: " & pkgInfo.basicInfo.name)
 
-proc getDownloadInfo*(pv: PkgTuple, options: Options,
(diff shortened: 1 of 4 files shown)


dom96 (changes requested):
Thank you for putting this together.

Two things:

* CI failing
* I think actually detecting whether the argument is an existing file would work quite well. Specifically restricted to .nimble files. No need for the --nimble-file flag.

PR #1023, opened 2022-08-24
Implement async download based on chronos

yyoncho:
- implementation of async download based on chronos asyncproc
- the PR is based on #938


FedericoCeratto:
Nimble is shipped with the Nim compiler and should not depend on external libraries.

Issue #482, opened 2018-04-24
Feature: Task-level dependencies


mratsim:
I'd like to use [Quickcheck](https://github.com/alehander42/nim-quicktest) for testing but it seems like currently either it's listed as a full dependency and get pulled on user machine, or it's not listed and it cannot be run as a nimble task.

Can we also allow `requires` at the task level.

dom96:
Perhaps you could run ``nimble install ...`` in the task?

mratsim:
I just tried, that unfortunately didn't work :/

```
  Executing task test_property_debug in /Users/<username>/Programming/Status/stint/stint.nimble
   Warning: Package 'stint' has an incorrect structure. The top level of the package source directory should contain at most one module, named 'stint.nim', but a file named 'io.nim' was found. This will be an error in the future.
      Hint: If this is the primary source file in the package, rename it to 'stint.nim'. If it's a source file required by the main module, or if it is one of several modules exposed by 'stint', then move it into a 'stint/' subdirectory. If it's a test file or otherwise not required to build thethe package 'stint.nim', prevent its installation by adding `skipFiles = @["io.nim"]` to the .nimble file. See https://github.com/nim-lang/nimble#libraries for more info.
Downloading https://github.com/alehander42/nim-quicktest using git
   Warning: Package 'quicktest' has an incorrect structure. The top level of the package source directory should contain at most one module, named 'quicktest.nim', but a file named 'js_lib.nim' was found. This will be an error in the future.
      Hint: If this is the primary source file in the package, rename it to 'quicktest.nim'. If it's a source file required by the main module, or if it is one of several modules exposed by 'quicktest', then move it into a 'quicktest/' subdirectory. If it's a test file or otherwise not required to build the the package 'quicktest.nim', prevent its installation by adding `skipFiles = @["js_lib.nim"]` to the .nimble file. See https://github.com/nim-lang/nimble#libraries for more info.
  Verifying dependencies for quicktest@0.18.0
      Info: Dependency on random@#head already satisfied
  Verifying dependencies for random@#head
      Info: Dependency on https://github.com/alehander42/breeze.git@>= 0.0.5 already satisfied
  Verifying dependencies for breeze@0.0.5
 Installing quicktest@0.18.0
    Prompt: quicktest@0.18.0 already exists. Overwrite? [y/N]
    Answer: y
   Success: quicktest installed successfully.
  Verifying dependencies for stint@0.0.1
[...]

PR #1035, opened 2022-11-04
Implement task level dependencies

ire4ever1190:
Implements #482

Syntax works like so
```nim
taskRequires "taskName", "somePackage == 0.1.0"
```

This is done instead of allowing `requires` at task level since it means a task such as `test` doesn't need to be redefined just to add dependencies to it (also is easier to implement).

More work still needs to be done, making draft now in case anyone wants to give input while I hack away at it.

What needs to be done (These might work, I just need to write test cases to see if they work)

- [x] Support lock files
- [x] Support `deps` related commands (All dependencies should be outputted)
- [x] Support develop

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -73,16 +73,13 @@ proc processFreeDependencies(pkgInfo: PackageInfo, options: Options):
   ##
   ## Returns set of PackageInfo (for paths) to pass to the compiler
   ## during build phase.
-
   assert not pkgInfo.isMinimal,
          "processFreeDependencies needs pkgInfo.requires"
   # If the folder we are operating in is the same folder that the .nimble
   # file is in then we know we are working the main package
   let
     isMain = options.startDir == pkgInfo.myPath.splitpath().head
-    command = case options.action.typ
-      of actionCustom: options.action.command.normalize
-      else: ""
+    task = options.task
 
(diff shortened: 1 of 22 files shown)


ire4ever1190:
Ready for review :+1:

src/nimble.nim
```diff
@@ -84,8 +88,27 @@ proc processFreeDependencies(pkgInfo: PackageInfo, options: Options):
...
+    ## Adds all task requirements to list of requirements
+    for task, requires in pkgInfo.taskRequires:
+      requirements &= requires
+
+  # Check what task level dependencies need to be added
+  if isMain:
+    if task in pkgInfo.taskRequires:
+      # If this is the main file then add its needed requirements for running a task.
+      requirements &= pkgInfo.taskRequires[task]
+    elif options.action.typ in {actionLock, actionSync}:
```
Araq, line 105: Why do these requirements (which are optional) have to be added to the lock file?

ire4ever1190: They are only added to the lock file when generating/syncing it so that if someone wants to run the task then it will use the same locked version. When loading the lock file the task dependencies are skipped unless running the task (So they stay optional)

Araq: Makes sense.

PR #1047, opened 2022-12-18
Fix lock file regression

ire4ever1190:
Fixes regression I caused in https://github.com/nim-lang/nimble/commit/c0f2f78dd06f67f99612afd45975b495057122fd where not all dependencies were loaded from a lock file.

Fixes it by storing task dependencies in different subgraphs so they can be loaded separately from the main dependency graph. The versions are still all resolved together though so the versions still need to be valid across tasks like before. The lock file now has extra key `tasks` which stores the extra dependencies.


zah:
All changes from this PR were rebased and incorporated in https://github.com/nim-lang/nimble/pull/1017 which is now merged and likely to be shipped in the next Nimble version (`0.15.0`).

PR #1054, opened 2022-12-29
replaces the redirect link with the direct link

xilicode:
fix packageLists


src/nimblepkg/config.nim
```diff
@@ -24,6 +24,7 @@ proc initConfig(): Config =
   result.cloneUsingHttps = true
   result.packageLists["official"] = PackageList(name: "Official", urls: @[
     "https://github.com/nim-lang/packages/raw/master/packages.json",
```
ringabout, line 26:
```suggestion
```
Replacing the redirect link with the direct link in order to save one redirect, which is quite important for users with poor/semi-broken internet.

PR #1063, opened 2023-01-16
Make sure that nimble.paths is checked relative to config.nims

yyoncho:
- this change will allow executing the

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -1857,17 +1857,17 @@ proc setupNimbleConfig(options: Options) =
     configFileVersion = 2
     sectionEnd = "# end Nimble config"
     sectionStart = "# begin Nimble config"
-    configFileHeader = &"# begin Nimble config (version {configFileVersion})\n"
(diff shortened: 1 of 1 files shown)


src/nimble.nim
```diff
@@ -1857,17 +1857,17 @@ proc setupNimbleConfig(options: Options) =
...
     configFileContentNoLock = fmt"""
 {configFileHeader}
-when system.fileExists("{nimblePathsFileName}"):
+when withDir(thisDir(), system.fileExists("{nimblePathsFileName}")):
   include "{nimblePathsFileName}"
 {sectionEnd}
 """
     configFileContentWithLock = fmt"""
 {configFileHeader}
 --noNimblePath
```
zah, line 1869: We've agreed that this should be conditional on the presence of the lockfile.

yyoncho: The reason for using lock file is that you can write the following...

```
nimble --lock-file=differentLockFileName.lock setup
```

... and it will still do the right thing. I don't have a strong preference for the approach here - both seem to be good enough.

Issue #1070, opened 2023-02-20
[Request] Tag a new release


yyoncho:
Given the important fixes that went in recently, it will be nice if a new release is tagged. Projects using lock files cannot be build with the latest tag: e. g. https://github.com/nim-lang/langserver/issues/34

ringabout:
Please change the version first.

ringabout:
The Nim language ships specific commit of the nimble, which doesn't relies on the tag of nimble. It might be able to ship a moderner nimble which includes the important fixes, though.

yyoncho:
PR at https://github.com/nim-lang/nimble/pull/1071

ringabout:
fixed

PR #1071, opened 2023-02-20
Bump version

yyoncho:
- see https://github.com/nim-lang/nimble/issues/1070


ringabout:
https://github.com/nim-lang/nimble/blob/a0a12f475905028763186ffb2341b92e2d8b2547/src/nimblepkg/common.nim#L25 needs to be changed too.

ringabout:
Thank you!

PR #1072, opened 2023-02-20
Force line endings when cloning

yyoncho:
- If we don't force them then depending on the target os the checksum might
fail.


Araq:
Shouldn't instead the checksum be made oblivious to the used line endings? But I don't what is better here, just pointing out an alternative.

yyoncho:
There is already `core.autocrlf=false` passed as a param so it seems like the decision to go in that direction has already been made.

Araq:
Ok.

PR #1073, opened 2023-02-23
Make the arguments style insensitive

yyoncho:
Before this PR there was a mix of cammel case and dash separated parameters. I
have unified the help using the dash version and also I have changed the
parameter handling to ignore the style to be backward compatible.

### DIFF
--- a/readme.markdown
+++ b/readme.markdown
@@ -221,7 +221,7 @@ See the [Creating Packages](#creating-packages) section for more info on this.
 
 Dependencies required for developing or testing a project can be installed by
 passing `--depsOnly` without specifying a package name. Nimble will then install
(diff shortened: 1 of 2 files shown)


Araq:
Dash separated parameters are not used by the Nim compiler though so that's quite unfortunate.

yyoncho:
> Dash separated parameters are not used by the Nim compiler though so that's quite unfortunate.

Then I can change the docs and examples to be camel case (dashes will still work). Let me know if you prefer going in this direction.

Araq:
> Then I can change the docs and examples to be camel case (dashes will still work). Let me know if you prefer going in this direction.

Oh yes, thanks!

PR #1076, opened 2023-03-04
 add support for -g flag, fix #1069

### LINKED ISSUE #1069: `nimble develop -g` doesn't work
```
options.nim(661)         handleUnknownFlags

    Error:  Unknown option: -g
```

https://github.com/nim-lang/nimble/blob/a0a12f475905028763186ffb2341b92e2d8b2547/src/nimblepkg/options.nim#L620

Adding "g" should suffice.

> -g, --global - Creates an old style link file in the special links directory. It is read by Nim to be able to use global develop mode packages. Nimble uses it as a global develop file if a local one does not exist.

foxoman:
The -g flag in the Nimble package manager would result in an error due to the flag being unrecognized, it is short of global flag


ringabout:
@foxoman Nice work!

PR #1090, opened 2023-04-14
Implement the command upgrade

yyoncho:
The command can be used to upgrade a minimal set of requirements in a lock file.

Here it is a sample invocation:

```
nimble upgrade foobar@#head
```

I have reworked the way `lock` operation works. Now it uses the lock file and
then applies the requirements from `upgrade`/`requires` section.


src/nimble.nim
```diff
@@ -639,22 +646,24 @@ proc processLockedDependencies(pkgInfo: PackageInfo, options: Options, onlyNim =
 
   let developModeDeps = getDevelopDependencies(pkgInfo, options)
 
+  var res: seq[PackageInfo]
```
zah, line 649: What's the benefit of having this intermediate sequence that will be converted to a hash table at the end of the function?

src/nimble.nim
```diff
@@ -639,22 +646,24 @@ proc processLockedDependencies(pkgInfo: PackageInfo, options: Options, onlyNim =
...
     elif isInstalled(name, dep, options):
-      result.incl getDependency(name, dep, options)
+      res.add getDependency(name, dep, options)
     elif not options.offline:
       let
         downloadResult = downloadDependency(name, dep, options)
-        dependencies = result.toSeq.filterIt(dep.dependencies.contains(it.name))
-      result.incl installDependency(pkgInfo.lockedDeps.lockedDepsFor(options).toSeq.toTable,
-                                    downloadResult, options, dependencies)
+        dependencies = res.toSeq.filterIt(dep.dependencies.contains(it.name))
```
zah, line 659: Since `res` is already a sequence here, the `toSeq` call is redundant.

yyoncho: Fixed.

src/nimble.nim
yyoncho: My idea was to remove the `toSeq` that I actually forgot to remove below. I went through several iterations to try to reduce these conversions in this method and in a few more places.

PR #1092, opened 2023-04-24
drop deps sha1

ringabout:
ref https://github.com/nim-lang/Nim/pull/21702

If nimble depends on `std/checksums`, it might make everything complex. Fortunately, only a small function needs to be duplicated.

### DIFF
--- a/src/nimblepkg/sha1hashes.nim
+++ b/src/nimblepkg/sha1hashes.nim
@@ -1,7 +1,7 @@
 # Copyright (C) Dominik Picheta. All rights reserved.
 # BSD License. Look at license.txt for more info.
 
-import strformat, strutils, json, std/sha1, hashes
+import strformat, strutils, json, hashes
 import common
 
 type
@@ -25,6 +25,10 @@ proc invalidSha1Hash(value: string): ref InvalidSha1HashError =
   ## Creates a new exception object for an invalid sha1 hash value.
   result = newNimbleError[InvalidSha1HashError](
     &"The string '{value}' does not represent a valid sha1 hash value.")
+
+proc isValidSha1Hash*(s: string): bool =
+  ## Checks if a string is a valid sha1 hash sum.
+  s.len == 40 and allCharsInSet(s, HexDigits)
 
 proc initSha1Hash*(value: string): Sha1Hash =
   ## Creates a new `Sha1Hash` object from a string by making all latin letters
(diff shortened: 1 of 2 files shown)


Araq:
Why close this? It's a good idea.

ringabout:
Nope, I was wrong. It actually depends on more procs of sha1.

Two ways out

- duplicate the sha1 module
- depends on the compiler via `nimHasChecksums`

```nim
when defined(nimHasChecksums):
  import "$compiler/dist/checksums/checksums/sha1"
else:
  import std/sha1

arnetheduck:
why would it be a problem for nimble to depend on a (locked version of a) nimble package? It already depends on Nim which is a "package".

arnetheduck:
basically, depending on a package turns an undeclared, unversioned and implicit dependency into a declared, versioned explicit dependency - the latter is less "complex" in the sense that there is more information available to resolve things that otherwise can go wrong.

ringabout:
Ideally, the compiler and the nimble should depend on the same version of sha1 while putting it at the deps meaning that two places need to be updated.

It probably also mean that the nimble needs to bootstrap itself from csource or something.

ringabout:
> All you need to do is to patch the build to git clone https://github.com/nim-lang/checksums.

The Nim compiler and the nimble [share the same checksums](https://github.com/nim-lang/nimble/pull/1094)?

Araq:
I don't understand the question. sha1 is sha1.

arnetheduck:
> Ideally, the compiler and the nimble should depend on the same version of sha1 while putting it at the deps meaning that two places need to be updated.

"the compiler" and "checksums" are two moving targets - they'll depend on the same version only iff this information is recorded somewhere: either via a lock file, a submodule or some other creative way of designating a particular version of each.

Since the two  projects (nim and nimble) are uncoupled, there's really no way to tightly express this constraint, that both must the depend on the exact same version, except if for example the compiler explicitly declares the version it uses and nimble reuses that.

Consider that nimble can be compiled with a "range" of nim versions (that ideally should be broad) - ditto checksums, nimble can specify a range of "checksums" packages that it's compatible with and because both implement sha1, this should be good enough (ie the difference between sha1 versions is mainly in the API - the outcome is defined by a standard so it doesn't greatly matter if compiler and nimble use the exact same version  / code).

PR #1100, opened 2023-04-26
disable `nimble build nimble` tests temporarily

ringabout:


ringabout:
Let me check `before` hook or something.

PR #1101, opened 2023-04-26
fixes nimble install nimble

ringabout:


ringabout:
Since tests cloned old nimble, so I have to merge the changes first.

PR #1105, opened 2023-05-03
Expose path of executing nimble to tasks

Gruruya:
So the executing nimble can be used in tasks i.e. `exec nimbleExe & " install -y"` over `exec "nimble install -y"`

### DIFF
--- a/src/nimblepkg/nimscriptapi.nim
+++ b/src/nimblepkg/nimscriptapi.nim
@@ -233,6 +233,8 @@ template after*(action: untyped, body: untyped): untyped =
     success = true
     retVal = `action After`()
 
+const nimbleExe* {.strdefine.} = "nimble"
+
 proc getPkgDir*(): string =
   ## Returns the package directory containing the .nimble file currently
   ## being evaluated.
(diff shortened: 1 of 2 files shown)


Gruruya:
Actually, the [--define:nimbleExe](https://github.com/nim-lang/nimble/pull/1105/commits/8dcd7b20afff43c6ad34d032caeb556e1bb1c32a) method is better, I'm going to revert back to that. It keeps `nimbleExe` in the scope of `nimscriptapi.nim`.

Gruruya:
Perhaps $PATH could be altered to make exec "nimble install -y" execute the current running nimble instead.

src/nimblepkg/nimscriptapi.nim
```diff
@@ -233,7 +233,7 @@ template after*(action: untyped, body: untyped): untyped =
     success = true
     retVal = `action After`()
 
-const nimbleExe* {.strdefine.} = "nimble"
+const nimbleBin* {.strdefine.} = "nimble"
```
Araq, line 236: I like `nimbleExe` much better, "bin" is some silly UNIX misnomer. It's a computer, everything is binary.

Gruruya: I don't really get "bin," but I changed it to match `nimBin` in `options.nim` https://github.com/nim-lang/nimble/blob/43a9fbffa4a28e7de07883adb09fc4a0bf4133f9/src/nimblepkg/options.nim#L44

Want to change `nimBin` to `nimExe`?

PR #1112, opened 2023-06-20
Add flag to allow skipping of the validation during nimble develop

yyoncho:


yyoncho:
@Araq the label didn't kick in after the rebase.

PR #1113, opened 2023-06-20
Add --mm:refc when building nimble

yyoncho:
- without this flag `nimble lock` crashes. Note that tests are also compiled
with this flag so do the nimble packaged from Nim repo


yyoncho:
@Araq 

@zah told me that this might be Nim bug. Let me know if you want me to report it in Nim repo.

Araq:
I'm only interested if `devel` still miscompiles it and then mostly in a reduced test case.

PR #1131, opened 2023-08-16
Improve version management of nim by nimble

yyoncho:
The following algorithm is implemented:

1. Use --nim param if specified(no validation of the version)
2. Use develop version if present(no validation of
the version)
3. Use locked nim
4. Use global version if matches `require` section
5. Download and install matching version of `requires` specification does not
match the existing one

### DIFF
--- a/tests/tnonim.nim
+++ b/tests/tnonim.nim
@@ -1,5 +1,5 @@
-# # Copyright (C) Dominik Picheta. All rights reserved.
-# # BSD License. Look at license.txt for more info.
+# Copyright (C) Dominik Picheta. All rights reserved.
+# BSD License. Look at license.txt for more info.
 
 {.used.}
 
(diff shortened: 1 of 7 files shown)


tests/tnonim.nim
```diff
@@ -1,12 +1,13 @@
-# Copyright (C) Dominik Picheta. All rights reserved.
-# BSD License. Look at license.txt for more info.
+# # Copyright (C) Dominik Picheta. All rights reserved.
+# # BSD License. Look at license.txt for more info.
```
tersec, line 2: `# # `?

yyoncho: will be reverted

yyoncho: Fixed.

arnetheduck:
lgtm - in future versions we might want to use a precompiled version and / or not compile tooling (which makes the build take much longer) but those are orthogonal concerns

arnetheduck (approved):

arnetheduck:
one side effect of the tooling being built is that another nimble is built too - depending on PATH settings, this may interfere with the nimble that the user is currently using - that's something to consider for commands that create shells

PR #1136, opened 2023-09-04
Don't ask questions when --silent flag is enabled; fixes #1066

### LINKED ISSUE #1066: nimble --silent demands answers to unknown questions
$ nimble --silent --localdeps install fusion
        Answer: 

and, as you can see above, this is not true:

       --silent                    Hide all Nimble and Nim output

.

    nimble v0.14.0 compiled at 2023-01-05 04:19:15
    git hash: couldn't determine git hash

xTrayambak:
- Don't ask questions when --silent is set; fixes #1066
- Remove accidental files

The bug in question (#1066) states that when `--silent` is passed over, it works as intended, until it comes to asking a question. In that case, it doesn't print the question, as intended, but still attempts to read the answer. This patch just makes all questions default to `true` if `--silent` is set so that no questions need to be asked.


xTrayambak:
Also, sorry for the first commit that added a ton of garbage. It definitely makes reviewing a nightmare but I forgot to remove `nimbledeps/` beforehand. Perhaps that directory should be added to `.gitignore`?

ringabout:
> Perhaps that directory should be added to .gitignore?

Make sense

Issue #1139, opened 2023-09-05
packages.json on GitHub and the Nim website are badly desynchronized.


xTrayambak:
This is in reference to issue #1129. It seems that https://nim-lang.org hosts waaaay less packages than the GitHub packages.json

Either the Nim hosting should be entirely removed or it should be synchronized with the GitHub one as many people (myself included) often times get served stale data when GitHub times out (happens for me at times).
![image](https://github.com/nim-lang/nimble/assets/59499552/e825cef7-94cd-4162-b94e-248bc195c643)

xTrayambak:
Most new Nimble packages *do not* exist in the Nim package "database". This leads to a ton of confusion. Perhaps when downloading the packages list, Nimble should atleast tell you where it's downloading them from.

Araq:
I had no idea we host it at all...

xTrayambak:
I didn't either, until I was scouring the code and found that Nimble tries to find packages from two URLs.

Issue #1129, opened 2023-08-07
Nimble downloading old packages.json


xjzh123:
Whenever Nimblle tries to download package index, it downloads an old file with 4220 lines and 303 items. The last package is:
```json
  {
    "name": "listsv",
    "url": "git://github.com/srwiley/listsv.git",
    "method": "git",
    "tags": [
      "singly linked list",
      "doubly linked list"
    ],
    "description": "Basic operations on singly and doubly linked lists.",
    "license": "MIT",
    "web": "https://github.com/srwiley/listsv"
  }
```

Nimbie version:
```
nimble v0.14.2 compiled at 2023-08-01 01:51:39
git hash: couldn't determine git hash
```

xTrayambak:
This is now fixed.

PR #1141, opened 2023-09-05
Temporarily remove outdated package list URL that was serving stale data

xTrayambak:
- remove outdated nimble package files

See #1139 and #1129 for more details.


src/nimblepkg/config.nim
```diff
@@ -23,8 +23,13 @@ proc initConfig(): Config =
   result.chcp = true
   result.cloneUsingHttps = true
   result.packageLists["official"] = PackageList(name: "Official", urls: @[
-    "https://raw.githubusercontent.com/nim-lang/packages/master/packages.json",
-    "https://nim-lang.org/nimble/packages.json"
```
ringabout, line 27: Though I would appreciate that "https://nim-lang.org/nimble/packages.json" is updated, which would greatly improve my user experience.

xTrayambak: Yep. GitHub sometimes stops working for me as well.

futureisfoss: I would also appreciate having "https://nim-lang.org/nimble/packages.json" as a fallback option because my connection to that github domain can be unreliable sometimes.

xTrayambak:
Shoot, I was counting each line like a package. Nope, 26K packages were NOT indeed gone, the number is probably smaller, but still significant.

ringabout:
@xTrayambak Hello, with the great help of @PMunch, https://nim-lang.org/nimble/packages.json is up to date! Eventually, there will be a script to update it automatically.

xTrayambak:
That's great to hear! I'll work on it tomorrow if that's fine.

ringabout:
No problem, thank you very much!

PR #1142, opened 2023-09-07
Don't use outside module nim-lang/checksums, use std/sha1 instead; fixes #1127

### LINKED ISSUE #1127: Cannot compile with nim 1.6
```
/nimble/src/nimblepkg/sha1hashes.nim(6, 42) Error: cannot open file: ../../dist/checksums/src/checksums/sha1
```

412af022a44157dd17118432fba825da9f204125

xTrayambak:
- Add more entries to .gitignore
- Remove dependency on nim-lang/checksums and use std/sha1 instead

#1127 states that sometimes the checksums module fails to get cloned, which is a real issue whilst compilation and in theory, *can* cause CI builds to fail. To prevent this, that library has been replaced with stdlib's `sha1` module which does exactly what the previous one does, except it is inside the standard library so Nimble no longer has to depend on another external library.


xTrayambak:
Wait, there's a tiny error. Let me fix it really quickly.

Issue #1158, opened 2023-11-08
ambiguous identifier: 'requiresData'


stoneface86:
Consider this nimble file
```nim
# Package

version       = "0.1.0"
author        = "stoneface86"
description   = "A new awesome nimble package"
license       = "MIT"
srcDir        = "src"


# Dependencies

requires "nim >= 1.6.16"

task echoRequires, "":
  echo requiresData

```

When attempting to access the `requiresData` var, an error occurs:
```
     Error: Could not read package info file in /mnt/ramdisk/ws/temp/temp.nimble;
        ...   Reading as ini file failed with: 
        ...     Invalid section: .
        ...   Evaluating as NimScript file failed with: 
        ...     /mnt/ramdisk/ws/temp/temp.nimble(14, 1) template/generic instantiation of `task` from here
        ... /mnt/ramdisk/ws/temp/temp.nimble(15, 8) Error: ambiguous identifier: 'requiresData' -- use one of the following:
        ...   system.requiresData: seq[string]
        ...   nimscriptapi_402557162.requiresData: seq[string]
        ... printPkgInfo() failed.
```

Using `system.requiresData` results in an empty seq, even though it should contain the nimble file's dependencies. When specifying it from the nimscriptapi_xxxxx module it gives us what we expect. I believe `requiresData` is missing from the except clause at https://github.com/nim-lang/nimble/blob/4fbb96d0dd05244f0ea2acfeb30b730f356aa38b/src/nimblepkg/nimscriptwrapper.nim#L108-L111

When I manually add `requiresData` to the generated nimscript file in `/tmp` it works

This occurs with nimble versions v0.13.1 and v0.14.2 on my machine.

ringabout:
I suppose it was fixed by https://github.com/nim-lang/nimble/pull/1159

PR #1159, opened 2023-11-09
fix issue #1158

stoneface86:
Fix for #1158. When accessing the `dataRequires` var, nim reports that it is an ambiguous identifier and a module is needed to specify which one (either `system` or the auto-generated `nimscript_xxxxxxxx`). By adding `dataRequires` to the except list in nimscriptwrapper.nim, this error is avoided and the var can now be used as intended.

I have also included a test case in tests/tissues.nim


Araq:
Excellent work, thank you!

Issue #1166, opened 2023-12-04
`compiler` dependency does not give access to compiler


arnetheduck:
.nimble:
```nim
requires "nim >= 1.6.16",
         "compiler"
```

```sh
nimble setup -l
   Warning: Using project local deps mode
      Info: using /home/arnetheduck/src/nimph/nimbledeps/pkgs2/nim-2.0.0-35b9d926d314ce8c3326902c138f5c2247cfd390/bin/nim for compilation
  Verifying dependencies for nimph@0.1.0
     Info:  "nimble.paths" is updated.
     Info:  "config.nims" is already set up.
...

nimble.paths:
```
arnetheduck@praeceps:~/src/nimph$ cat nimble.paths
--noNimblePath
```

notably, compiler path is missing (`compiler/` is not part of `nim` standard paths)

daylinmorgan:
Related to #1163

jmgomez:
Should we treat it as a some sort of "special" package? As far as I can tell is defined as:

```json
  {
    "name": "compiler",
    "alias": "nim"
  },
  ```
  
 Without knowing much about nimble, all I can say is either we expand the definition or we treat `compiler` differently

daylinmorgan:
Nimble is translating that to the `nim` package [nimpkgs link instead of packages.json](https://github.com/nimpkgs/nimpkgs/blob/main/packages/nim.json).

Which when using either `nimble install nim` or `nimble install compiler` is using a `before install` task defined in the [`nim.nimble`](https://github.com/nim-lang/Nim/blob/devel/nim.nimble) in the main repo to run the build script. But ignores it as `requires "compiler"` in a `.nimble`.

Being able to access the "compiler" as a source itself should really only be necessary if you need a version separate from the compiler you are using. Since imports can reference the compiler with `import "$nim" / compiler`.

We could either package the compiler source as a standalone repo that would be importable with `import compiler` and it's own `compiler.nimble` that doesn't build the nim binary or tools. Or modify `nimble` to do this behind the scenes and treat the `compiler` dependency special rather than as an alias for `nim`. 

~~I would say you we could do something with `https://github.com/nim-lang/nim/subdir?=compiler` and put the `compiler.nimble` there~~ Oh wait this already exists!

But the current `compiler.nimble` isn't usable with `nimble`, since `nimble` only copies the subdir to the `nimbleDir`. 

But if we change it to `include "$lib"` instead of `include ../lib`. Then the following seems to work for using the compiler as a package.
[...]

PR #1185, opened 2024-02-17
Adds the `paths` feature option

jmgomez:
#https://github.com/nim-lang/nimble/issues/1166


tests/tester.nim
```diff
@@ -1,33 +1,33 @@
...
+# import truncommand
 import tsetupcommand
-import ttestcommand
-import ttwobinaryversions
-import tuninstall
-import ttaskdeps
+# import ttestcommand
+# import ttwobinaryversions
+# import tuninstall
+# import ttaskdeps
```
Araq, line 30: Why were these tests disabled?

jmgomez: Apologies, I committed the file by mistake. Reverted it.

readme.markdown
```diff
@@ -1103,6 +1103,7 @@ Nimble includes a ``publish`` command which does this for you automatically.
   listed in ``bin``. Possible values include: ``c``, ``cc``, ``cpp``, ``objc``,
   ``js``.
   **Default**: c
+* ``paths`` - A list of relative paths that will be expanded on `nimble.paths` and the search paths options to the compiler.
```
arnetheduck, line 1106: these would always be relative to the nimble file?

arnetheduck: can they be used to escape the package sandbox? ie if I do `../../../etc`, can I get access to `/etc/passwd` with `staticRead` in a malicious package?

jmgomez: Doesnt `staticRead` uses the `appDir` as root? 

This PR would just add another directory to the compiler search path, which you could previously access anyways.
For instance, previously the directory would be something like: 

`/Volumes/Store/Projects/testclient/nimbledeps/pkgs2/nim-2.1.1-a4e01d628f14da66e92b8d2d41d7c0ec48c3bb05`

If you have `paths = @["compiler"]` it will get expanded to: 

```
--path:"/Volumes/Store/Projects/testclient/nimbledeps/pkgs2/nim-2.1.1-a4e01d628f14da66e92b8d2d41d7c0ec48c3bb05"
--path:"/Volumes/Store/Projects/testclient/nimbledeps/pkgs2/nim-2.1.1-a4e01d628f14da66e92b8d2d41d7c0ec48c3bb05/compiler"
```

arnetheduck: aka what would `paths = @["../../compiler"]` expand  to?

src/nimble.nim
```diff
@@ -355,6 +356,18 @@ proc allDependencies(pkgInfo: PackageInfo, options: Options): HashSet[PackageInf
   for requires in pkgInfo.taskRequires.values:
     result.incl pkgInfo.processFreeDependencies(requires, options)
 
+proc isSubdirOf(subdir, baseDir: string): bool =
+  subDir.normalizedPath.startsWith(baseDir.normalizedPath)
```
Araq, line 360: This is only correct if normalizedPath ends in `/`!

jmgomez: I dont understand, `normalizedPath` get rids of `/` i.e. `/a/b/c/` becomes `/a/b/c`

PR #1191, opened 2024-03-06
Add `nimble add` command

xTrayambak:
This PR adds the `nimble add` command, similar to `cargo add`. \
It verifies that the package does actually exist in the Nimble package index, unless it is a URL. \
It also verifies that the package isn't already a dependency to the current working directory's Nimble file.


arnetheduck:
Something this command should likely do is add a minimum version dependency on the current version of the package, ie if chronos 4.0 is available at the time of `add`, the added dependency should be `chronos >= 4.0.0` (or possibly its semver equivalent, ie `~4.0.0`)

xTrayambak:
Unfortunately `~/.nimble/packages_<xyz>.json`'s schema does not contain the current version, so we'll have to clone the package, install it, and then fetch the version. I'll look into that later.

xTrayambak:
Nevermind, I found an easy way to do this.

PR #1200, opened 2024-03-28
SAT implementation 

jmgomez:
Disabled tests will be added in a future PR

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,7 @@ jobs:
         run: nim r src/nimblepkg/private/clone.nim
       - name: Run nim c -r tester
         run: |
+          nimble install
           cd tests
           nim c -r tester
           # there's no need to add nimblepkg unit tests --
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -32,6 +32,7 @@ jobs:
         run: nimble install -y
       - name: Run nim c -r tester
         run: |
+          nimble install
```
ringabout, line 35:
```suggestion
```

src/nimble.nim
```diff
@@ -8,17 +8,17 @@ import std/options as std_opt
 
 import strutils except toLower
 from unicode import toLower
-
+import sat/sat
```
ringabout, line 11: Now or later, you need to ensure `koch` bundles `nimble` correctly, which needs to clone `sat` instead.

.github/workflows/test.yml
jmgomez: It needs `nimble install` for running the tests, no?

ringabout: > run: nimble install -y

My previous PR had added it already above

PR #1209, opened 2024-04-10
Enables SAT by default to test against the CI

jmgomez:


xTrayambak:
Compiling Nimble right now gives this error:
```
/home/tray/Documents/Projects/nimble/src/nimble.nim(12, 29) Error: cannot open file: ../dist/sat/src/sat/sat
       Tip: 6 messages have been suppressed, use --verbose to show them.
nimble.nim:229           buildFromDir

    Error:  Build failed for the package: nimble
```
Is the SAT solver done or are you still working on it? I'd be down to helping out if you need any help :)

jmgomez:
> Compiling Nimble right now gives this error:
> 
> ```
> /home/tray/Documents/Projects/nimble/src/nimblepkg/nimblesat.nim(337, 35) Error: attempting to call undeclared routine: 'addUnique'
>        Tip: 6 messages have been suppressed, use --verbose to show them.
> nimble.nim:229           buildFromDir
> 
>     Error:  Build failed for the package: nimble
> ```

The error is because that function is only in newer std versions. Will take care of it. 

> Is the SAT solver done or are you still working on it?

Im at the point of making it work with all other Nimble commands and it needs real world testing before making it the default, which probably should be aligned with making local install the default too. The cache also needs more work.

src/nimble.nim
```diff
@@ -51,14 +51,22 @@ proc checkSatisfied(options: Options, dependencies: seq[PackageInfo]) =
           [pkgInfo.basicInfo.name, $currentVer, $pkgsInPath[pkgInfo.basicInfo.name]])
     pkgsInPath[pkgInfo.basicInfo.name] = currentVer
 
+proc displaySatiesfiedMsg(pkgs: HashSet[PackageInfo]) =
```
xTrayambak, line 54: Perhaps rename this to `displaySatisfiedMsg`?

xTrayambak (approved):

jmgomez:
Getting closer. Missing one test from tuninstall that only fails in linux and the lock feature

PR #1221, opened 2024-05-20
adds nimdir to `dump`

jmgomez:


jmgomez:
Do not merge it yet, Im going to make it part of `dump` so it can be easily cached inside the lsp as `dump` is already in use there and is taking quite a bit of time.

PR #1222, opened 2024-05-26
Add forge aliases for different Git forge providers

xTrayambak:
This PR adds "forge aliases" - they're an easy way to avoid writing out full URLs for packages that aren't on the Nimble index. \
They work similarly to how Nix's system works.

Here's a few examples:
```
$ nimble install gh:xTrayambak/librng
$ nimble install srht:bptato/chame
```

It also handles special "quirks" in a forge (eg., SourceHut adds a tilde in front of the username, so this handles it if the user didn't add it themselves) \
I've written a few tests for the alias "parser" as well. It works with the `install` and `add` commands, and as a result also works fine if you add the alias inside your
Nimble file like this:
```
# Package

version       = "0.1.0"
author        = "Lorem Ipsum"
description   = "Lorem Ipsum Door Sit"
license       = "MIT"
bin           = @["lorem"]

# Dependencies

requires "nim >= 2.0.0"
requires "gh:xTrayambak/librng"
requires "srht:bptato/chame"
requires "srht:~user/repository"
```
Please let me know if I missed something out or if I should add something else to this.

### DIFF
--- a/src/nimblepkg/forge_aliases.nim
+++ b/src/nimblepkg/forge_aliases.nim
@@ -1,0 +1,120 @@
+# Copyright (C) Dominik Picheta. All rights reserved.
+# BSD License. Look at license.txt for more info.
+
+import std/strutils
+import common
+
+type
+  ForgeKind* = enum
+    fgGitHub
+    fgGitLab
+    fgSourceHut
+    fgCodeberg
+
+  Forge* = ref object
+    case kind*: ForgeKind
+    of fgGitHub:
+      ghUsername*, ghRepo*: string
+    of fgGitLab:
+      glUsername*, glRepo*: string
+    of fgSourceHut:
+      shUsername*, shRepo*: string
+    of fgCodeberg:
+    cbUsername*, cbRepo*: string
+
+proc expand*(alias: Forge): string {.inline.} =
+  var expanded = "https://" # add an option to use http instead?
+
+  case alias.kind
+  of fgGitHub:
+    expanded &= "github.com/" & alias.ghUsername & '/' & alias.ghRepo
+  of fgGitLab:
+    expanded &= "gitlab.com/" & alias.glUsername & '/' & alias.glRepo
+  of fgSourceHut:
+    expanded &= "git.sr.ht/" & alias.shUsername & '/' & alias.shRepo
+  of fgCodeberg:
+    expanded &= "codeberg.org/" & alias.cbRepo & '/' & alias.cbRepo
+
+  expanded
+
+proc isForgeAlias*(value: string): bool {.inline.} =
+  let splitted = value.split(':')
+
+  splitted.len == 2
+
+proc parseForgeKind*(value: string): ForgeKind {.inline.} =
+  let splitted = value.split(':')
+  
+  case splitted[0].toLowerAscii()
+  of "github", "gh": return fgGitHub
+  of "gitlab", "gl": return fgGitLab
+  of "sourcehut", "srht", "shart": return fgSourceHut
+  of "codeberg", "cb", "cberg": return fgCodeberg
+  else:
+    raise nimbleError("Invalid forge alias name: " & value[0])
+
+proc parseGenericAlias*(value: string, appendTilde: bool = false): tuple[username, repo: string] {.inline.} =
+  let splitted = value.split(':')
+
+  if splitted[1].len < 1:
+    raise nimbleError(
+      "Invalid forge alias format; correct format:" & 
+      "\n\t<alias>:<username>/<repository>\n" &
+      "     ^^^^^"
+    )
+
+  let secondSplit = splitted[1].split('/')
+
+  if secondSplit.len < 2:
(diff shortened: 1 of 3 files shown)


src/nimblepkg/forge_aliases.nim
```diff
@@ -0,0 +1,120 @@
...
+  splitted.len == 2
+
+proc parseForgeKind*(value: string): ForgeKind {.inline.} =
+  let splitted = value.split(':')
+  
+  case splitted[0].toLowerAscii()
+  of "github", "gh": return fgGitHub
+  of "gitlab", "gl": return fgGitLab
+  of "sourcehut", "srht", "shart": return fgSourceHut
+  of "codeberg", "cb", "cberg": return fgCodeberg
```
jtbx, line 52: Codeberg is one of many [forgejo](https://forgejo.org/) instances, perhaps extract Codeberg's logic out into a new "forgejo" type, allowing e.g. `forgejo:git.example.org:foo/bar`. It would most likely work for legacy Gitea instances too. You could do the same for sourcehut but selfhosted sourcehut instances are much rarer.

src/nimblepkg/forge_aliases.nim
```diff
@@ -0,0 +1,120 @@
...
+
+  splitted.len == 2
+
+proc parseForgeKind*(value: string): ForgeKind {.inline.} =
+  let splitted = value.split(':')
+  
+  case splitted[0].toLowerAscii()
+  of "github", "gh": return fgGitHub
+  of "gitlab", "gl": return fgGitLab
+  of "sourcehut", "srht", "shart": return fgSourceHut
```
jtbx, line 51: Looks good :ok_hand:

jtbx (approved):

src/nimblepkg/forge_aliases.nim
xTrayambak: Yeah, but these are for forge providers, not forge services. You might as well put in the full URL at that point.

xTrayambak:
I'm gonna try to fix the failing CI really quickly, sorry for that one.

src/nimblepkg/forge_aliases.nim
```diff
@@ -0,0 +1,120 @@
...
+  Forge* = ref object
+    case kind*: ForgeKind
+    of fgGitHub:
+      ghUsername*, ghRepo*: string
+    of fgGitLab:
+      glUsername*, glRepo*: string
+    of fgSourceHut:
+      shUsername*, shRepo*: string
+    of fgCodeberg:
+    cbUsername*, cbRepo*: string
```
Araq, line 23:
```suggestion
      cbUsername*, cbRepo*: string
```

src/nimblepkg/forge_aliases.nim
```diff
@@ -0,0 +1,120 @@
...
+  Forge* = ref object
+    case kind*: ForgeKind
+    of fgGitHub:
+      ghUsername*, ghRepo*: string
+    of fgGitLab:
+      glUsername*, glRepo*: string
+    of fgSourceHut:
+      shUsername*, shRepo*: string
+    of fgCodeberg:
+      cbUsername*, cbRepo*: string
```
Araq, line 23:
```suggestion
type
  ForgeKind* = enum
    fgGitHub
    fgGitLab
    fgSourceHut
    fgCodeberg

  Forge* = object
    kind*: ForgeKind
    username*, repo*: string
```

There seems to be no advantage to require the user of the `Forge` type to do the case analysis.

xTrayambak: Yeah, seems like it. I was originally planning to add self-hosted forges into this as well (as @jtbx suggested) but I later realized it'd be counter intuitive. I'll fix this.

jmgomez:
@xTrayambak do you mind to attempt to pass the CI with sat on?

xTrayambak:
I'm not sure how the SAT solver works, sorry.

jmgomez:
@xTrayambak np, to enable it is just `useSatSolver*: bool = true` in `options.nim`. I just did

xTrayambak:
We're not enabling it yet though, right?

jmgomez:
Nope, we need to test it in the real world. But it's a good idea to make sure the CI keeps passing for it when meaningful changes are merged

jmgomez:
SAT is still green :)

src/nimblepkg/forge_aliases.nim
```diff
@@ -0,0 +1,93 @@
...
+import common, download
+
+type
+  ForgeKind* = enum
+    fgGitHub
+    fgGitLab
+    fgSourceHut
+    fgCodeberg
+
+  Forge* = ref object
```
Araq, line 14: Why is that a `ref`? It doesn't look like it needs reference semantics.

xTrayambak: Ah, I'd assumed that objects with field discriminators need to be `ref`s.

PR #1225, opened 2024-06-04
Import compiler from "$nim"

ehmry:


PMunch:
This is currently blocking `nim-lang/choosenim` from being able to compile on newer versions of Nim.

PR #1228, opened 2024-06-18
Implements entryPoints opt-in.

jmgomez:
A list of relative paths to nim files that will be used by the `nimlangserver` as project entry points.


Araq:
This needs to be documented.

PR #1233, opened 2024-07-01
Improves nim installation by using csources (same as atlas)

### LINKED ISSUE #1175: Remove nimble from nim compilation
As of https://github.com/nim-lang/nimble/pull/1131, `nimble` can build nim - however, as part of building nim-the-compiler, it also builds a toolchain including `nimble` - this can lead to the user instance of nimble being replaced with the version that gets compiled with nim.

when `nimble` installs `nim` as part of the dependencies of a project, it should not compile `nimble` and other tooling as this makes `nimble shell` and similar behave inconsistently as `nimble` gets replaced by a different version.

jmgomez:
Notice when SAT is enabled, csources is stored in the cache. 
Missing:
- [x] Proper reporting
- [x] Handle special versions (assume csources_2?). We could have special cases for #version-x
- [x] Refactor to another file (nimble.nim is crazy big already)
- [x] Test cases
- [x] Fixes an issue where the packageCache wasnt being shared in local mode
- [x] Remove nimble from nim compilation Fixes #1175

### DIFF
--- a/src/nimblepkg/nimenv.nim
+++ b/src/nimblepkg/nimenv.nim
@@ -77,6 +77,8 @@ proc compileNim*(options: Options, nimDest: string, v: VersionRange) =
     if not keepCsources:
       removeDir workspace / csourcesVersion / "c_code"
     let pathEntry = workspace / nimDest / "bin"
+    #remove nimble so it doesnt interfer with the current one:
+    removeFile "bin" / "nimble".addFileExt(ExeExt)
     when defined(windows):
       writeFile "activate.bat", BatchFile % pathEntry.replace('/', '\\')
     else:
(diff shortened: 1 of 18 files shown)


src/nimblepkg/nimenv.nim
```diff
@@ -0,0 +1,110 @@
...
+  var major, minor, patch: int
+  if not nimVersion.isSpecial: 
+    if not scanf($nimVersion, "$i.$i.$i", major, minor, patch):
+      display("Error", "cannot parse version requirement", Error)
+      return
+  let csourcesVersion =
+    #TODO We could test special against the special versionn-x branch to get the right csources
+    if nimVersion.isSpecial or (major == 1 and minor >= 9) or major >= 2:
+      # already uses csources_v2
+      "csources_v2"
```
Araq, line 48: More future proof would be:

```nim

if major >= 3: "csources_v" &$major

```

But it's also somewhat risky to predict the future this far.

jmgomez: ok but It's an easy enough change to do when `csources` is increased as there is a test already that uses it in the `nimble` ci

Issue #144, opened 2015-06-14
Have Nimble verify that a package does not install modules outside its namespace


dom96:
Make sure conventions are followed.

N.B. Make sure to disallow a 'pkg' suffix *for package names*

dom96:
Verify package names too https://github.com/nim-lang/packages/issues/361#issuecomment-224769791

PR #1236, opened 2024-07-06
Remove incorrect package structure warnings.

planetis-m:
As discussed https://forum.nim-lang.org/t/11942 and probably in more places before. This PR removes all checks for package structure, these are two but they seem closely tied, regardless everything should work fine even without the warnings.


src/nimblepkg/packageparser.nim
```diff
@@ -176,10 +117,6 @@ proc validatePackageInfo(pkgInfo: PackageInfo, options: Options) =
     if pkgInfo.backend notin ["c", "cc", "objc", "cpp", "js"]:
       raise validationError("'" & pkgInfo.backend &
           "' is an invalid backend.", false)
-  if options.action.typ in {actionInstall, actionBuild, actionDevelop, actionCompile, actionCheck}:
```
planetis-m, line 179: Alternatively we can limit the warning to show for only actionCheck, actionDevelop.

planetis-m:
Fails at installing different Nim versions tniminstall.nim seems unrelated.

jmgomez:
> Fails at installing different Nim versions tniminstall.nim seems unrelated.

`tniminstall.nim` doesnt fail. What fails is `  [FAILED] can validate package structure (#144)` in `tissues.nim`

PR #1254, opened 2024-08-04
(fix) packageinfo: checkInstallFile now correctly compares files; fixes #1226

### LINKED ISSUE #1226: skipFiles not working
`skipFiles` has no effect with `installExt` or `installDirs`.

### Sample package structure,

```
folder
|- app.nimble
|- src
    |- app.nim
    |- appPkg
        |- utils.nim
        |- protocol.nim
```

### Corresponding nimble file

```nimble
srcDir        = "src"
installExt    = @["nim"]
skipFiles     = @["appPkg/utils.nim"]
```

### Expected Behaviour

All files inside `appPkg` should be copied except `appPkg/utils.nim`

### Actual Behaviour

`appPkg` is copied as it is.

xTrayambak:
`checkInstallFile` incorrectly calls `samePath` to compare a file from the install list and a file from the ignore list. This PR just makes the function compare the file from the install list to the absolute path of the file from the ignore list. This fixes #1226

### DIFF
--- a/src/nimblepkg/packageinfo.nim
+++ b/src/nimblepkg/packageinfo.nim
@@ -411,7 +411,8 @@ proc checkInstallFile(pkgInfo: PackageInfo,
   for ignoreFile in pkgInfo.skipFiles:
     if ignoreFile.endswith("nimble"):
       raise nimbleError(ignoreFile & " must be installed.")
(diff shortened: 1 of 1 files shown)


xTrayambak:
@Araq @jmgomez Could either of ya review this?

PR #1257, opened 2024-08-10
Improvenimdirselection

jmgomez:
## returns the nim directory prioritizing the nimBin one if it satisfais the requirement of the project
  ## otherwise it returns the major version of the nim installed packages that satisfies the requirement of the project
  ## if no nim package satisfies the requirement of the project it returns the nimBin parent directory
  ## only used by the `nimble dump` command which is used to drive the lsp

### DIFF
--- a/src/nimblepkg/nimenv.nim
+++ b/src/nimblepkg/nimenv.nim
@@ -1,4 +1,4 @@
-import std/[strscans, os, strutils, strformat]
+import std/[strscans, os, osproc, strutils, strformat, options]
 import version, nimblesat, cli, common, options
 
 when defined(windows):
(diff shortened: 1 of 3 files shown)


src/nimblepkg/nimenv.nim
```diff
@@ -120,3 +120,13 @@ proc useNimFromDir*(options: var Options, realDir: string, v: VersionRange, tryC
...
+
+proc getNimVersionFromBin*(nimBin: string): Option[Version] =
+  let cmd = nimBin & " --version"
+  if nimBin.fileExists:
+    let info = execProcess(cmd)
+    var major, minor, patch: int
+    for line in info.splitLines:
+      if scanf(line, "Nim Compiler Version $i.$i.$i", major, minor, patch):
+          let ver = $major & "." & $minor & "." & $patch
+          return some newVersion(ver)
```
Araq, line 132:
```suggestion
        let ver = $major & "." & $minor & "." & $patch
        return some newVersion(ver)
```

PR #1258, opened 2024-08-14
Add `sponsor` command

xTrayambak:
# Prelude

[See this issue](https://github.com/nim-lang/packages/issues/2917) \
This PR exists as a base for what can be the final implementation for this. Do **NOT** merge this yet! \
This PR adds a new command: `nimble sponsor`

# The sponsor command
The sponsor command optionally takes in a package as an argument. If one is not provided, it shows all packages that are accepting donations (the package below does not actually take donations, it's just been modified to show how this command works in my packages_official.json)
image

If a package is provided, the links for sending donations to the library maintainer are opened in the user's browser and also outputted to their terminal if they don't have a web browser installed. (Again, the package below does not actually take donations.)
image

### DIFF
--- a/src/nimblepkg/packageinfo.nim
+++ b/src/nimblepkg/packageinfo.nim
@@ -3,7 +3,7 @@
 
 # Stdlib imports
 import system except TResult
-import hashes, json, strutils, os, sets, tables, times, httpclient, strformat
+import hashes, json, strutils, os, sets, tables, uri, times, httpclient, strformat
 from net import SslError
 
 # Local imports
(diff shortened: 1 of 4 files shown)


src/nimblepkg/packageinfo.nim
```diff
@@ -18,6 +18,21 @@ proc initPackageInfo*(): PackageInfo =
 proc initPackage*(): Package =
   result = Package(version: notSetVersion)
 
+proc `$`*(meth: DonationMethod): string {.inline.} =
+  ## Turns a `DonationMethod` into a pretty string.
+  case meth
+  of DonationMethod.GitHub: "GitHub"
+  of DonationMethod.OpenCollective: "OpenCollective"
+  of DonationMethod.Patreon: "Patreon"
```
juancarlospaco, line 26: I think this should be just a string metadata instead of just 3 hardcoded "provider companies",
some people prefer to use USDT/BTC/ETH/crypto/other services etc, 
and in some places of the world the hardcoded services are not available (as donations collector).

xTrayambak: Sounds alright, but we'd also need to properly audit any new packages that are added in. We need to ensure that none of them end up linking to a phishing site, albeit the odds are very low.

juancarlospaco: Thats what I tried to do at
https://github.com/juancarlospaco/nim_packages_security_audit?tab=readme-ov-file#fully-automated-nim-packages-security-audit
Feel free to contribute/revive, maybe it needs like a "cool web visualization" (?). 

Also the package scanner can be improved too:
https://github.com/nim-lang/packages/blob/master/package_scanner.nim

I think the efforts there are so much better spent.

xTrayambak: This might sound stupid - but what about using something like URL Haus, VirusTotal or Google Safebrowsing to the package scanner?

xTrayambak: Nevermind, URL Haus does not keep track of phishing URLs (my main concern).

xTrayambak: @juancarlospaco I'd say we should leave the URL auditing to the person who's going to merge the package into the index. Most browsers have SafeBrowsing built in either ways. Do you have any other suggestions?

juancarlospaco: Enforce a common-sense standard max len for the string ? 256 char or something here or in the package scanner. IMHO 🤔

PR #1260, opened 2024-08-21
Removes unnecessary dep satisfied logging. Fixes an issue where the c…

jmgomez:
…ache wasnt being properly set. Dont re-download packages if they already exists

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -97,7 +97,7 @@ proc processFreeDependenciesSAT(rootPkgInfo: PackageInfo, options: Options): Has
 
   result = solveLocalPackages(rootPkgInfo, pkgList, solvedPkgs)
   if solvedPkgs.len > 0: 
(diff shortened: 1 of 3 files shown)


src/nimble.nim
```diff
@@ -97,7 +97,7 @@ proc processFreeDependenciesSAT(rootPkgInfo: PackageInfo, options: Options): Has
 
   result = solveLocalPackages(rootPkgInfo, pkgList, solvedPkgs)
   if solvedPkgs.len > 0: 
-    displaySatisfiedMsg(solvedPkgs, pkgsToInstall)
+    # displaySatisfiedMsg(solvedPkgs, pkgsToInstall)
```
arnetheduck, line 100: we can simply switch this to debug level where it does make sense to log "positive" progress

jmgomez: Will review and rework if necessary the logging system in another pr (test may start to fail if we change how overall logging works)

PR #1266, opened 2024-09-26
Adds `requires` flag which allows to add extra packages to the dependency resolution

jmgomez:


src/nimble.nim
```diff
@@ -153,12 +155,14 @@ proc processFreeDependencies(pkgInfo: PackageInfo,
   ## during build phase.
   assert not pkgInfo.isMinimal,
          "processFreeDependencies needs pkgInfo.requires"
-
+  var requirements = requirements
   var pkgList {.global.}: seq[PackageInfo]
   once: 
     pkgList = initPkgList(pkgInfo, options)
-    if options.useSatSolver:
+    if options.useSatSolver:
```
Araq, line 162:
```suggestion
    if options.useSatSolver:
```

PR #1279, opened 2024-10-28
Fixes #1277 Tagged versions (i.e. #head) behaves like `any` except when matched against another tag

### LINKED ISSUE #1277: SAT resolution failure
In the [nim-libp2p](https://github.com/status-im/nim-libp2p) repository, dependency resolution is failing when using the SAT dependency solver.

These are the scenarios:
1. `nimble --solver:legacy install`
    - Works
2. `nimble --solver:sat install`
    - Fails
3. `nimble install_pinned` (a custom mechanism that installs fixed versions from a `.pinned` file into a local `nimbledeps` directory, which is recreated each time it's run)
    - The outcome varies depending on how it's run:
        a) When run with legacy (`nimble --solver:legacy install_pinned`), it works.
        b) When run with SAT (`nimble --solver:sat install_pinned`), it fails.
        c) Interestingly, if you first run it with legacy and then run `nimble --solver:sat check`, the check passes.

At first I thought this was due to SAT not being able to resolve the dependencies in `libp2p.nimble`, but scenario `3c` made me think that's not quite it.

I'm not sure anymore if this happens because our dependencies are actually unsolvable or there's a situation SAT just can't handle. Any ideas?

jmgomez:
Fixes #1277

### DIFF
--- a/src/nimblepkg/version.nim
+++ b/src/nimblepkg/version.nim
@@ -141,8 +141,8 @@ proc withinRange*(ver: Version, ran: VersionRange): bool =
     return ver <= ran.ver
   of verEq:
     return ver == ran.ver
-  of verSpecial:
-    return ver == ran.spe
+  of verSpecial: # special version is always within range
+    return ver.isSpecial and ver == ran.spe or not ver.isSpecial
   of verIntersect, verTilde, verCaret:
     return withinRange(ver, ran.verILeft) and withinRange(ver, ran.verIRight)
   of verAny:
--- a/src/nimblepkg/version.nim
+++ b/src/nimblepkg/version.nim
@@ -65,6 +65,9 @@ proc initFromJson*(dst: var Version, jsonNode: JsonNode, jsonPath: var string) =
 proc isSpecial*(ver: Version): bool =
   return ($ver).len > 0 and ($ver)[0] == '#'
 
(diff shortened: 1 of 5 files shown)


src/nimblepkg/version.nim
```diff
@@ -141,8 +144,8 @@ proc withinRange*(ver: Version, ran: VersionRange): bool =
     return ver <= ran.ver
   of verEq:
     return ver == ran.ver
-  of verSpecial:
-    return ver == ran.spe
+  of verSpecial: # special version is always within range
+    return ver.isSpecial and ver == ran.spe or not ver.isSpecial
```
arnetheduck, line 148: btw is this the tag name or the resolved hash?

jmgomez: afaict its whatever follows `#`i.e. `#head` `#branch` `#commit`

arnetheduck: I think the ideal here is that the actual resolved version is used - that exploits de-facto compatibility between two branches, should they happen to point to the same commit

jmgomez: `withinRange` is too fundamental, it needs to be fast and be ready to do the lookups quite early on. But we can explore the idea at some point I guess. Right now there is a huge ordering issue that needs to be addressed with a major refactor

PR #1296, opened 2024-12-01
Fix forge aliases with SAT solver

xTrayambak:
The SAT solver seems to have borked the forge alias logic. This PR fixes it.

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -789,7 +789,7 @@ proc install(packages: seq[PkgTuple], options: Options,
         if not isAlias:
           getDownloadInfo(pv, options, doPrompt) #TODO dont download if its nim
         else:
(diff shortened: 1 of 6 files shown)


jmgomez:
@xTrayambak thanks. Can you add a test case?

xTrayambak:
@jmgomez Sorry, I'm not quite sure what to write a test-case for here. I've never really worked with the Nimble SAT solver, so I can't do that.

jmgomez:
@xTrayambak you dont need to test anything specific to SAT, just an integration tests where you run `nimble install` on a package that uses the `pv.name.isForgeAlias` code path

xTrayambak:
Alright. I'll wait for this one to run.

PR #1301, opened 2024-12-05
Fixes CI: Update binaries macos

jmgomez:


.github/workflows/binaries.yml
```diff
@@ -42,7 +42,7 @@ jobs:
             builder: ubuntu-20.04
           - target:
               os: macosx
-            builder: macos-12
+            builder: macos-14
```
arnetheduck, line 45: from what I remember of github action runners, this will switch it to arm64

jmgomez: Ah good to know, we will need to ship universal then

PR #1302, opened 2024-12-06
Update binaries to produce macos x86

jmgomez:


arnetheduck:
another one of these thing1s that would do great as a common nimble github action: producing binaries from nimble packages, for lots of platforms

PR #1314, opened 2024-12-16
UX and Test Improvements

elcritch:
There's a few different UX improvements in this PR: progress spinner, colorized and prettified `nimble deps`, and an inverted deps format `nimble deps --format=inverted`.

## Progress Spinner 

Nimble can take a while to check git for package caches now among other things, which can lead to frustration not knowing if Nimble is working or hung on some repo.

This PR adds a spinner style progress indicator to make these waits less annoying. The progress is based on suppressed debug messages which gives a good cadence to the work being done.

The spinner is just a simple `    Working  ⡿` style one, not a percentage one. It only shows when using `noColor` is not set.

image

## Deps Tree

It's colorized and outputs a nice tree format borrowing from Cargo, NPM, etc.

image

## Inverted Deps Tree

It's *really* helpful to print each dependency for a project and what requirements led to the resolved version.

image

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -42,7 +42,7 @@ jobs:
             builder: ubuntu-20.04
           - target:
               os: macosx
-            builder: macos-12
+            builder: macos-14
           - target:
               os: windows
             builder: windows-2019
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -102,6 +102,7 @@ jobs:
       - name: Download artefacts
         uses: actions/download-artifact@v4
 
+      # Create/update the "latest" release
       - uses: ncipollo/release-action@v1
         with:
           name: Latest Nimble Binaries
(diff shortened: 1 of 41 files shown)


arnetheduck:
This is nice - nimble is way way way way too verbose and fills the screen with useless junk for basically anything it does - it should not be wasting vertical space on "I actually managed to perform the task you asked me to" as if this was a surprise, every time.

> The spinner is just a simple

instead of "working", it could show what it's doing (ie `Scanning xyz` or `Cloning ...`)

arnetheduck:
> some line matching

this sounds like a maintenance nightmare though - maybe better leave it as is until a more deliberate source of information can be developed

elcritch:
> this sounds like a maintenance nightmare though - maybe better leave it as is until a more deliberate source of information can be developed

It could be. It shouldn't result in anything breaking just going to the default "Working". 

I added it as I was curious and it doesn't seem _too_ bad. Does make it nicer to know it's "Scanning" (printPkgInfo) or "Updating" (git ls-remote).

Though the idea is nice, I'll probably revert it. It'd be better and probably not too hard to pipe the info in via the `doCmd` pieces.

elcritch:
@jmgomez this is ready. Not sure who's reviewing these things nowadays so figured I'd ping you.

jmgomez:
Looks good to me. But @Araq is the one approving for the time being :)

PR #1326, opened 2024-12-30
Only caches nim release for a day

jmgomez:


Araq:
But why?

jmgomez:
> But why?

Because when a new release is out it will be out of the cached filed so it will install a previous one

Araq:
We dont release on a daily basis though.

jmgomez:
I feel like a request per day is a good compromise, the whole point of the cache was not to prevent the request for happening but to dont reach the GH API limit. 
But we can change it if you have a better number for the frequency

PR #1328, opened 2025-01-09
Cli ux improvements docs

elcritch:
Ooops, forgot to update some of the help docs for my last PR. The new commands aren't helpful if folks can't find them!

Also decided to change `nimble publish --tags` to `nimble publishTags`. It helps split out options for the commands and their options, while still grouping similar actions for discoverability.


arnetheduck:
`nimble publishTags` feels wrong / too much though. `nimble tag` would make more sense, where it would read the nimble file and put a corresponding git tag and do nothing else (ie not push the tag to any remotes). in particular, changing past tags is a big no-no and the commit that changes the nimble file is not always the desired release version.

PR #1330, opened 2025-01-10
UX Change to show checksums and git commits in deps output

elcritch:
Useful for seeing what version and checksum combo is actually used. 

Example output. 

image


elcritch:
Note the double windy package shouldn't be there, but it's useful to at least see that it's there.

PR #1337, opened 2025-01-23
Introduces an experimental declarative parser into nimble

jmgomez:


jmgomez:
Replaced by: https://github.com/nim-lang/nimble/pull/1342

PR #1345, opened 2025-02-11
Use package info for nimble path when available

### LINKED ISSUE #1344: nimble path does not respect current package configuration
When running `nimble path [pkgname]`, it doesn't respect the configuration of the current package. It ignores the `develop` file and it ignores any specific version configuration in the nimble file.

My expectation is that when I run `nimble path` from a nimble directory, it should print the path of the dependency as contextualized by the configuration in the current package.

Nycto:
This change updates the nimble path command so it respects the local package configuration when executed within a nimble package. Fixes #1344


jmgomez:
You need to rebase for the MacOS test to pass

jmgomez:
There are failing tests

PR #1351, opened 2025-02-27
Features can be activated through dependencies

jmgomez:


src/nimble.nim
```diff
@@ -80,6 +80,23 @@ proc addReverseDeps(solvedPkgs: seq[SolvedPackage], allPkgsInfo: seq[PackageInfo
...
+    if pkg.isNone: 
+      displayError &"PackageInfo {solved.pkgName} not found", priority = LowPriority
+      continue
+    if pkg.get.activeFeatures.len == 0:      
+      pkg = some pkg.get.toRequiresInfo(options)
+    for pkgTuple, activeFeatures in pkg.get.activeFeatures:
+      let pkgWithFeature = getPackageInfo(pkgTuple[0], allPkgsInfo, none(Version))
+      if pkgWithFeature.isNone:
+        displayError &"Active PackageInfo {pkgTuple[0]} not found", priority = HighPriority
+        continue
```
Araq, line 97: If I read the code correctly, this `continue` doesn't do anything.

jmgomez: Right, good catch. AI autocomplete gotchas

jmgomez: Well no, it does. I missed the next line. It will skip it. Although a break would be better so it doesnt keep trying

PR #1352, opened 2025-02-28
Fix bash completion errors

micl2e2:
When being used in conjunction with `nim`'s completion scripts in such style:
```
source /path/to/nim.bash-completion
source /path/to/nimble.bash-completion
```

, and there would be an error while completing `nim`:
```
<TAB>
bash: ilastlast: substring expression < 0
```

`nim`'s completion fails because both share the same helper functions: `__is_short_or_long`, `__ask_for_subcmd_or_subopts`, `__ask_for_subcmd`, `__ask_for_subcmd_opts`, and the impl of `nimble`'s ones are a little bit more outdated than `nim`'s ones, and multiple `source`s will overide the previous definition. This pr simply reuses those functions in `nim`, `nimgrep`, etc.

### DIFF
--- a/nimble.bash-completion
+++ b/nimble.bash-completion
@@ -1,5 +1,9 @@
 # bash completion for nim                             -*- shell-script -*-
 
+# (NOTE) The following functions SHOULD be identical to those
+# in nim source tree: __is_short_or_long, __ask_for_subcmd_or_subopts,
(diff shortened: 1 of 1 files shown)


jmgomez:
> This pr simply reuses those functions in nim, nimgrep, etc.

Does it assume that nimble is released with nim?

PR #1365, opened 2025-03-12
implements `namedBin` for the declarative parser

jmgomez:


src/nimblepkg/declarativeparser.nim
```diff
@@ -56,6 +56,24 @@ proc extractFeatures(featureNode: PNode, conf: ConfigRef, hasErrors: var bool):
         extractRequires(stmt, conf, requires, hasErrors)
         result.add requires
 
+proc extractTableLiteral(n: PNode, conf: ConfigRef): Table[string, string] =
+  ## Extracts a table literal of the form {"key": "value"} or {"key": "value"}.toTable()
```
Araq, line 60: No, that's terrible. Instead only allow:

```nim
namedBin      = {
  "features_deps": "features_deps"
}
```

jmgomez: Thats allowed too, but not allowing `toTable()` would break when parsing the file with the vm

Araq: How so? is `namedBin` an existing feature? Never heard of it...

jmgomez: Yes, it is. i.e. nimbus uses it https://github.com/status-im/nimbus-eth2/blob/fad7b9a7133faf63efb552b6caaa3fb1554aba19/beacon_chain.nimble#L56

Araq: Does the existing Nimble require the `toTable`? If not, don't support it and make the Nimbus people update their nimble file.

arnetheduck: the question would then be what it should be updated _to_ - the feature is used to build binaries whose "main" files are in different folders which is something that eventually happens to any project of sufficient complexity.

Araq: Update to a version that lacks the `toTable` call. A single line change.

jmgomez: > Does the existing Nimble require the `toTable`?

Yes, otherwise it wont be a table anymore.. Also notice we still require the vm parser not just for retrocompatibility (in case the new one fails) but also some of the features (i.e. tasks) are still retrieved using the vm one

Araq: Then you need to bite the bullet and also support the prefix `toTable({....})` syntax.

arnetheduck: wasn't the deterministic parser supposed to focus on version matters only? ie requires and version and nothing else?

once a version  has been selected (of nim in particular, but also of everything else), the nimble file can be passed through the vm and "full execution" stuff like this can continue to work..

broadly, that's probably the way forward here in the sense that we can bring more and more stuff over to "declarative parsing" which tooling can reason about and leave the rest for the VM to process. it would also be interesting to mark the "side-effectness" of stuff executed in the vm to know what can be cached and what can't, eventually.

arnetheduck: > ship the vm with Nimble instead and remove the circular dep from the compiler

there's no circular dep though .. we parse the `requires` information declaratively - this gives us information about which nim to use to interpret the file (and what other packages we have to install).

we put those packages in place _and then_ we interpret the file using that nim and the output of said interpretation is the input to the next "nimble" processing phase, whatever that might be (running a task etc). this order of things allows the right nim version to be used, but it also ensures that when interpreting the nimble file, we can use code from the dependencies to resolve imports etc. including a vm would mean we're on the wrong version of the vm for that processing, sometimes.

jmgomez: > there's no circular dep though .

There is `nimble` specific code in the `nim compiler` no?

arnetheduck: > There is nimble specific code in the nim compiler no?

there's an agreement between the two to look in some nimble folders, but I think the best thing that can happen is that this code is removed in favor of nimble declaring to nim (via `nimble setup`) what it should do. ie whatever is there, is fairly contained and minimal

Araq: It took a month long painstaking process to rip Nimble off the Nim compiler dependency for it used to import the Nim compiler as a library for Nimble evaluation. We won't go back to that.

jmgomez: Didnt know that but as you know the compiler is used in the new parser anyways. Right now we have the bad parts of both worlds: the evaluation happens with an arbitrary `nim` different than the selected which we cant control. Maybe when the declarative parser is the new default we could workaround it by first selecting `nim`

Araq: Using the parser is a whole different story than using Nim's VM. The parser's API is stable and small, the VM's is messy and it keeps us from replacing it with a better one.

PR #1433, opened 2025-07-28
Adds an additional sat pass to hard require the current system Nim.

jmgomez:
This will make dependencies to adjust to the system nim version. If the resolution to fails, it relaxes the nim requires silently.


jmgomez:
Unrelated CI failure in MacOS

PR #1436, opened 2025-07-30
Fixes an issue where using url requires with special versions, sometimes ended with duplicate packages in the solved set

jmgomez:


src/nimblepkg/nimblesat.nim
```diff
@@ -821,14 +821,47 @@ proc normalizeRequirements*(pkgVersionTable: var Table[string, PackageVersions])
   # if pkgVersionTable.hasKey("json_serialization"):
   #   echo "DEBUG NEW VERSIONS FOR ", "json_serialization", " ", pkgVersionTable["json_serialization"].versions.mapIt(it.version).join(", ")
 
+proc clearSolvedPkgs(solvedPkgs: var seq[SolvedPackage]) =
```
arnetheduck, line 824: shouldn't making the name canonical happen before running any solvers on them? ie shouldn't this be part of loading the existing versions of a package, keeping in mind that the canonical name ostensibly can change over time?

jmgomez: Ideally yes but it requires a larger refactor that surely breaks `--legacy` thats why I did the post processing first but it may worth to spend the time now? What needs to be done is conceptually is to modify the "requires" for all packages so instead of an url they use the package name. Practically, the best way to go about it is to have an url field in the PackageMinimalInfo type

jmgomez: Well, the time to resolve it, its right after "collection" because its when you first have the information. When we are collecting the packages, what happens is we transform the package into a PackageMinimalVersion (that SAT can understand) and we lost the package info information afterwards. (In a later stage we get back the package either from memory or disk) 

```nim

  PackageMinimalInfo* = object
    name*: string
    version*: Version
    requires*: seq[PkgTuple]
    isRoot*: bool

  PackageVersions* = object
    pkgName*: string
    versions*: seq[PackageMinimalInfo]
    
 ```
 So basically after collecting all requires we end up with a `Table[string, PackageVersions]`.
 Then we feed it into the solver:
 
 ```nim
   collectAllVersions(pkgVersionTable, root, options, downloadMinimalPackage, pkgList.mapIt(it.getMinimalInfo(options)))
  solvedPkgs = pkgVersionTable.getSolvedPackages(output).topologicalSort()
  ```
  
 By storing the url as part of the minimal info we can do lookups to get the right package for a given require.

arnetheduck: > When we are collecting the packages

what I don't get is why the translation cannot be done there and then, so that `PMV` already contains the translated name. Presumably, "collection" means doing the git dance with the given url and at that stage, we have everything we need to do the translation as well.

Anyway, the most important point is to do it before sat runs so that correct data is fed _into_ sat - garbage in, garbage out as they say, and this is what we want to avoid.

Issue #1449, opened 2025-08-15
Road to Nimble 1.0.0


jmgomez:
# Nimble 1.0.0 Overview

The new **Nimble 1.0.0** introduces a SAT-based design with a brand-new greenfield code path, reducing complexity and improving maintainability.  
`local` dependency resolution will be the new default.

## Key Changes

- **No Nim required initially** – Resolves the Nim version from `compilation.nim`; no VM fallback for Nim itself.  
- **Centralized, upfront resolution** – Solves dependencies globally at the start, replacing the legacy greedy approach.  
- **Single-pass by default** – Uses a declarative parser for most cases; falls back to the VM parser (with the resolved Nim) only for tricky packages or `require` statements inside control flow.  
- **Local dependencies by default** – Ensures predictable behavior and avoids global pollution.  
- **Ensure compatibility with fallback** – Guarantees backward compatibility with existing projects.  
- **Lay the groundwork for a new choosenim alternative** – Provides the foundation for future Nim version management improvements.  
- **Improve speed** – Declarative parser is faster than the VM parser, along with faster Nim selection and other optimizations.  
- **Introduce “prefer” for Nim** – Uses the system Nim version if possible, selecting older dependency versions when needed.  
  *Example:* If system Nim is 2.0 but the latest package B requires 2.2, Nimble will choose an earlier package B that works with 2.0.  
- **Normalize dependencies** – Resolves package URLs to names before solving so different URLs for the same package aren’t treated as separate dependencies.

## Milestones

- [x] Implement upfront SATPass (main goal; required major refactor)  
- [x] Remove requirement for Nim at startup  
- [x] Make local dependencies the default  
- [ ] Improve speed
   - [x] Add benchmarks
   - [x] Optimize code paths
   - [ ] Enable parallel downloads
   - [x] Expand cache  
- [x] Ensure full backward compatibility with `legacy` flag (preserves old behavior)  
- [x] Implement dependency normalization  
- [x] Add `prefer` system Nim feature  
- [x] Support file paths for local `require`s  
- [ ] Improve `nimble develop` workflow  
- [x] Add Nim version management (e.g. `nimble install nim@2.0.8`)  
- [ ] Complete miscellaneous tasks: 
     - [x] Update `nimble dump
     - [x] nimble update nimble
     - [x] Battle test Nim selection
- [ ] Triage issue tracker
- [ ] Cleanup old codepath (`legacy`) entirely
- [ ] Implement `--minVer`
[...]

PR #1452, opened 2025-08-18
WIP Support for filepath in requires

jmgomez:
Implements "Support file paths for local requires" from https://github.com/nim-lang/nimble/issues/1449

- "Support file paths for local requires" is meant to be a replacement for `nimble develop`'s workflows (support for nimble develop is kept). 
- Filepaths can only be used in the current top level package being developer, or in a package that is required via `filepath`. 
- `filepaths` packages overwrites the SAT solution for a given package. i.e. if a dep has `results == 0.2.0` and there is a filepath pointing to a fork of `results` (i.e. `file://../results`) the latter will be used. 
- `filepaths` can be added in the nimble file or in a `requires` file near to the project (so they can be ignored).

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -2113,7 +2113,11 @@ proc lock(options: var Options) =
       if solvedPkg.pkgName.isNim and not shouldAddNim: continue
       
       # Get the PackageInfo for this solved package
-      let pkgInfo = options.satResult.getPkgInfoFromSolved(solvedPkg, options)
+      let pkgInfo = 
+        if solvedPkg.pkgName.isFileURL:
+          getPkgInfoFromDirWithDeclarativeParser(solvedPkg.pkgName.extractFilePathFromURL(), options)
+        else:
+          options.satResult.getPkgInfoFromSolved(solvedPkg, options)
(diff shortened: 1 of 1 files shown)


arnetheduck:
this must be blocked for remote packages or things could get messy when a package is published to git - ie it should only work for local packages - if a package/version was downloaded from git, it should be marked invalid if it contains such urls

jmgomez:
@arnetheduck 

>this must be blocked for remote packages or things could get messy when a package is published to git

Yeah, thought about it but its too late when the package is already in git. 

> the alternative is to not allow this at all and instead rely on develop

`develop` is so convoluted and has so many issues that 99% users just dont use it. We can fix the issues if we spot them, but the UX is there..

What about this design instead: we already have the special `feature "dev"` (short form is `dev:`) that it only gets enabled in the local package. We can make it so the declarative parser only allows `file://` requires inside the `dev` feature, outside of it would fail. If any other package has it in `dev` it wont be activated as it isnt the current package worked on. 

We can also preprocess the requirements in a way where we prefer the `file:://` package regardless the version if present. 

The downside is that we lost transitive `file:://` dependencies, but I dont think they are needed for this workflow.

PR #1457, opened 2025-08-28
Adds support some `when` expressions in the declarative parser.

jmgomez:


arnetheduck:
Overall, this approach represents a direction that we might ultimately not want in the future, ie whose details will likely change - backwards compatibility with packages that use `when` for now is already served by the fallback mode (which itself has some improvements scheduled) so there is no urgency to introduce partial compatibility features based on hacks like `compileDefines` just for the sake of enabling the parsing of a few more packages to use the new parser.

It's probably reasonable to keep this PR in draft until things have settled around other in-development things like feature flag set arithmetic - once that is done, we can revisit the "language" for conditional requirements and its interaction with the system / nim etc

Araq:
Atlas uses this approach successfully (and it doesn't offer a fallback at all). It's reasonably easy to make it sound, document the available defines and the fact that you have the `not`/`and`/`or` boolean operators. Produce an error for an unsupported define. It's not a hack, it merely started as one...

arnetheduck:
> reasonably easy to make it sound

thanks to the fallback in nimble, we can make it sound first, then add it as a feature, so that we don't have to maintain backwards compatibility with a potentially unsound or incomplete version in the new code) - ie apart from soundness, it would be good to have completeness as well in the same package and consider other formats for it (`requires(..., cond="win or mac"` for example before committing) once feature flags reach maturity since the two things are codependent, UX-wise.

PR #1459, opened 2025-09-02
Fixes an issue where packages with binDir couldnt being installed globally

jmgomez:


jmgomez:
Mac CI failures are unrelated "couldnt resolve host"

Issue #1486, opened 2025-09-26
`nimble deps` redundantly repeats the same deps over and over


arnetheduck:
```
nora
  nim >= 2.0.12
  web3 >= 0.8.0(resolved 0.8.0)
    nim >= 2.0.0
    chronicles@any(resolved 0.12.2)
      nim >= 2.0.16
      faststreams >= 0.3.0(resolved 0.5.0)
        nim >= 1.6.0
        stew >= 0.2.0(resolved 0.4.1)
          nim >= 1.6.0
          results@any(resolved 0.5.1)
            nim >= 1.2
          unittest2@any(resolved 0.2.4)
            nim >= 1.6.0
        unittest2@any(resolved 0.2.4)
          nim >= 1.6.0
      serialization@any(resolved 0.4.9)
        nim >= 1.6.0
        faststreams@any(resolved 0.5.0)
          nim >= 1.6.0
          stew >= 0.2.0(resolved 0.4.1)
            nim >= 1.6.0
            results@any(resolved 0.5.1)
              nim >= 1.2
            unittest2@any(resolved 0.2.4)
              nim >= 1.6.0
          unittest2@any(resolved 0.2.4)
            nim >= 1.6.0
```

Here we can see two packages depending on stew - both of them redundantly print the transitive deps of stew which are always the same - instead, it could just print `...` for the second appearance - as is, the list easily become thousands of lines long for even moderate projects - the above package for example, even though the number of unique packages it depends on is limited, fills several screenfuls with repeated deps.

Issue #1331, opened 2025-01-13
nimble reevaluates every package in nimble dir


SolitudeSF:
nimble reevaluates every package in nimble dir. if your global nimble dir is polluted, it makes a simple `nimble build` take multiple minutes to run.

SolitudeSF:
is package cache supposed to be stored in tmp? issue would also not be present, if it used persistent cache location, so nimble didn't need to reevaluate everythign after every reboot, but i have no clue, if package cache is supposed to be invalidated regularly.

jmgomez:
> is package cache supposed to be stored in tmp? 
Only for the `legacy` solver and in some edge cases. There is a `pkgcache` folder but thats orthogonal to the reason of why it takes that long. We cant cache `sat` just yet as the way `nimble requires` works today isnt deterministic with a cache in front. But we are aiming into that direction

PR #1718, opened 2026-05-28
Cache parsed package info in nimblemeta.json to avoid re-parsing inst…

bung87:
…alled packages

This fixes the performance issue where nimble reevaluates every package in the nimble directory on each command, causing slow builds when the global nimble dir has many packages (nim-lang/nimble#1331). same to https://github.com/nim-lang/nimble/issues/1486

Changes:
- Extend PackageMetaData with requires, features, srcDir, paths, preHooks, postHooks, and nestedRequires fields
- Save parsed info to nimblemeta.json during installation
- Load cached info in getInstalledPackageMin to skip re-parsing
- Skip toRequiresInfo for installed packages that have cached data
- Add backward-compatible initFromJson for PackageMetaData

### DIFF
--- a/src/nimblepkg/declarativeparser.nim
+++ b/src/nimblepkg/declarativeparser.nim
@@ -683,6 +683,10 @@ proc toRequiresInfo*(pkgInfo: PackageInfo, options: Options, nimBin: Option[stri
     display("Warning", msg, Warning, HighPriority)
     raise newNimbleError[BabelPackageError](msg)
 
+  # Skip re-parsing installed packages that have cached requires in metadata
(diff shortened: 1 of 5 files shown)


jmgomez:
The declarative parsing shouldnt take enough time to worth to cache it. Did you benchmark it?

bung87:
I didn't get correct performance result due to something wrong when running, but I have 200+ packages installed, so even the declarative parsing is fast, I still dont wanna waste time

jmgomez:
Well, you should use `local` which will be the default and has a cache already in place no need for yet another one

bung87:
I only use couple times due to network condition, most times use the global mode.

PR #1782, opened 2026-07-20
Make `nimble dump` return nimDir for nim#head and other "special" versions

### LINKED ISSUE #1781: `nimble dump`: nimDir is empty a special version of Nim is required
# Steps
1. `nimble init foo`
2. `cd foo`
3. Set required nim version in .nimble to `nim#head`.
4. `nimble install -ly`
5. `nimble setup`
6. `nimble dump`

# Expectation

nimDir points to the installed nim in nimbledeps.

# Reality

nimDir is empty.

moigagoo:
Fixes #1781

### DIFF
--- a/src/nimble.nim
+++ b/src/nimble.nim
@@ -2372,9 +2372,7 @@ proc getNimDir(options: var Options, nimBin: var Option[string]): string =
     try:
       solvePkgs(rootPackage, options, nimBin)
     except CatchableError:
-      # Issue #1713: dump never raises on resolution failure — it returns ""
-      # so the langserver can detect "couldn't pick a nim"
-      return ""
+      return nimBin.getNimBin.parentDir
     if nimBin.isNone:
       return ""
     return nimBin.getNimBin.parentDir
--- a/src/nimblepkg/versiondiscovery.nim
+++ b/src/nimblepkg/versiondiscovery.nim
@@ -452,6 +452,14 @@ proc fillPackageTableFromPreferred*(packages: var Table[string, PackageVersions]
         packages[pkg.name] = PackageVersions(pkgName: pkg.name, versions: @[pkg])
       else:
         packages[pkg.name].versions.add pkg
+    elif pkg.version.isSpecial and pkg.version.speSemanticVersion.isSome:
+      # Prefer the entry that carries a semantic version (derived from the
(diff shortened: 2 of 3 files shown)


moigagoo:
I think I found another bug that the fix won't cover:

1. `nimble init foo && cd foo`
2. Add `nim#head` to reqs.
3. `rm -rf nimble.paths nimbledeps/ config.nims && nimble install -ly && nimble setup && nimble dump`. nimDir points correctly to nimbledeps.
4. Add `unittest2` req.
5. `rm -rf nimble.paths nimbledeps/ config.nims && nimble install -ly && nimble setup && nimble dump`. nimDir points incorrectly to ~/.nimble.

src/nimble.nim
```diff
@@ -2372,9 +2372,7 @@ proc getNimDir(options: var Options, nimBin: var Option[string]): string =
     try:
       solvePkgs(rootPackage, options, nimBin)
     except CatchableError:
-      # Issue #1713: dump never raises on resolution failure — it returns ""
-      # so the langserver can detect "couldn't pick a nim"
-      return ""
+      return nimBin.getNimBin.parentDir
```
jmgomez, line 2375: I think this is not the right thing to do here. An empty nim should tell the langserver that resolution failed and it must ask the user to install the deps. Otherwise nothing would work, even you have "a compatible nim"
To remediate it
  - Keep return "" for genuine resolution failures.
  - Fix selection so the installed special Nim is actually resolved.
  - Add an integration test reproducing the issue exactly: local install, setup, then dump,
    asserting nimDir is inside nimbledeps.

I can tackle it for you (I dont mind as is likely less work on my side than reviewing), this may got deeper of what it seems

jmgomez (changes requested):

src/nimble.nim
moigagoo: @jmgomez If it is no too much of an inconvenience to you, I'd be happy to accept your offer, thanks 🙏

However, I can't fully understand why this is wrong. "Keep return "" for genuine resolution failures"—isn't this what happens on the line above?

```nim
if nimBin.isNone:
  return ""
```

jmgomez: No, nim can be set via the bootstrap mechanism as soon as the vm parser is required during enumeration

PR #1843, opened 2026-09-05
Implement feature and dev blocks for VM package parsing

elcritch:
## Summary

NimScript `feature` and `dev` blocks currently discard their bodies, so the VM parser cannot collect feature-scoped dependencies. This implements feature metadata collection for direct VM package parsing and for declarative parsing that falls back to the VM.

This complements #1836, which fixes dependency feature-list parsing in the declarative parser and SAT resolution.

## Implementation

- evaluate `feature` bodies only during the VM `printPkgInfo` metadata action
- route `requires` calls into the currently evaluated feature, including the `dev` shorthand
- retain named features with no dependencies
- emit feature-scoped dependencies in a `[Features]` metadata section and parse them into `PackageInfo.features`
- preserve VM-computed features after declarative-parser fallback instead of overwriting them with the partial declarative result
- update the `--features` help text and `PackageInfo.features` documentation to no longer describe features as declarative-only

Feature bodies remain inactive during normal Nimble actions.

## Test coverage

The VM regression belongs in `tnimscript.nim` and exercises both direct `getPkgInfoVm` parsing and the unified declarative-to-VM fallback with:

- a computed package version that forces VM fallback
- a computed feature dependency
- a conditional dependency inside a feature block
- an empty named feature
- a `dev` feature dependency

## Testing

- `cd tests && nim c -r tnimscript.nim`
- `cd tests && nim c -d:release -r tnimscript.nim "nimscript::collects feature blocks as VM metadata and through fallback"`


jmgomez:
Thanks for the PR. Half of it was already done in another PR. Willing to accept it but look at the AI feedback: 
 
Built this branch and ran it against a fixture — this regresses packages that take the VM fallback:

```nim
const packageVersion = "0.1.0"
version = packageVersion
feature "gui":
  requires "stew >= 0.2.0[featA, featB]"
```

`nimble --features:gui install` now yields `requires: @[stew@>= 0.2.0, featB]@any version, ...]` and fails on the bogus `featB]` package. master resolves it correctly.

Cause: `[Features]` is written with `join(", ")` and read back with `multiSplit`, which splits on every comma, and the new `if not result.isNimScript` guard makes that mangled result replace the declarative one.

Please follow the model the declarative parser already uses for features (`NimbleFileInfo.features: Table[string, seq[string]]`, one requirement per element, never flattened into a comma-joined string) instead of introducing a second, lossy representation on the VM side.

elcritch:
Addressed in ba65b4c. VM feature metadata now emits one `[Features]` entry per requirement, and the reader appends repeated entries instead of passing a comma-flattened value through `multiSplit`. Empty named features remain preserved. The regression test now uses `stew >= 0.2.0[featA, featB]` and checks both direct VM parsing and declarative-to-VM fallback. I also tested a temporary merge with #1836: the full `tnimscript` suite, the release-mode VM regression, and #1836’s focused dependency-feature parser and version-discovery tests pass.

PR #1844, opened 2026-09-05
Implement feature and dev blocks for VM package parsing

elcritch:
## Summary

NimScript `feature` and `dev` blocks previously discarded their bodies, so feature-scoped dependencies were available only from the declarative parser. This implements feature metadata collection in the VM parser and keeps it intact when declarative parsing falls back to VM evaluation.

## Implementation

- evaluate `feature` bodies only during the VM `printPkgInfo` metadata action
- route `requires` calls into the currently evaluated feature, including the `dev` shorthand
- retain named features with no dependencies
- emit feature-scoped dependencies in a `[Features]` metadata section and parse them into `PackageInfo.features`
- preserve VM-computed features after declarative-parser fallback instead of overwriting them with the partial declarative result
- update the `--features` help text and `PackageInfo.features` documentation to no longer describe features as declarative-only

Feature bodies remain inactive during normal Nimble actions.

## Test coverage

The VM-specific regression is in `tnimscript.nim`. It exercises both direct `getPkgInfoVm` parsing and the unified declarative-to-VM fallback with:

- a computed package version that forces VM fallback
- a computed feature dependency
- a conditional dependency inside a feature block
- an empty named feature
- a `dev` feature dependency

The existing declarative feature test is also run as a regression check.

## Testing

- `cd tests && nim c -r tnimscript.nim`
- `cd tests && nim c -d:release -r tnimscript.nim "nimscript::collects feature blocks as VM metadata and through fallback"`
- `cd tests && nim c -r tdeclarativeparser.nim "Declarative parser features::should be able to parse features from a nimble file"`


elcritch:
Superseded by #1843, which now includes both the dependency feature parsing fix and the VM feature-block implementation.

PR #1859, opened 2026-09-14
Make `nimble dump` read-only: never install Nim (fixes #1857)

### LINKED ISSUE #1857: `nimble dump` installs nim when `~/.nimble/pkgs2` is empty
From https://github.com/nim-lang/nimble/issues/1713 my understanding is `nim dump` should be read-only. However, it seems to install nim *at least* when the global ~/.nimble/pkgs2 is empty (which is common in CI and when using nimbledeps mode only, ig):

```
mkdir -p /tmp/dumprepro/proj/src
echo 'version       = "0.1.0"
author        = "repro"
description   = "repro"
license       = "MIT"
srcDir        = "src"
requires "nim >= 2.0.8"' > /tmp/dumprepro/proj/proj.nimble
echo "hi" > /tmp/dumprepro/proj/src/proj.nim
cd /tmp/dumprepro/proj
HOME=/tmp/dumprepro nimble -y dump    # downloads and install nim

ls /tmp/dumprepro/.nimble/pkgs2
ls /tmp/dumprepro/.nimble/nimbinaries
du -sh /tmp/dumprepro/.nimble
```

```
$ nimble --version
nimble v0.24.1 compiled at 2026-09-12 06:05:20
git hash: a1b471d13d173897942f99d013f4efb79913bdf2
```

Aside, `nim dump` should support being called concurrently; say spawned by different processes (or the same multiple times), particularly by MCP and LSP in langserver, they may conflict with each other if they are allowed to install Nim (leaving a broken nim version in pkgs2) or do other conflicting writes.

jmgomez:


tests/tnimbledump.nim
```diff
@@ -303,3 +303,80 @@ testEntryPoint: "tests/tall.nim"
...
+      putEnv("PATH", fakeBin)
+      defer:
+        putEnv("HOME", oldHome)
+        putEnv("PATH", oldPath)
+        if hadConfigHome:
+          putEnv("XDG_CONFIG_HOME", oldConfigHome)
+        else:
+          delEnv("XDG_CONFIG_HOME")
+
+      # --offline makes a regression fail fast instead of downloading a toolchain.
```
nitely, line 366: mmh, but it also risks requiring --offline to get this to work as it should...
or maybe both modes should be tested...

PR #1872, opened 2026-09-22
Write the same config.nims section whether or not there is a lock file

jmgomez:


src/nimble.nim
```diff
@@ -1903,26 +1903,18 @@ proc setupNimbleConfig(options: Options, nimBin: Option[string]) =
...
+    # `--noNimblePath`.
+    configFileContent = fmt"""
 {configFileHeader}
 when withDir(thisDir(), system.fileExists("{nimblePathsFileName}")):
   include "{nimblePathsFileName}"
 {sectionEnd}
-"""
-    configFileContentWithLock = fmt"""
-{configFileHeader}
---noNimblePath
```
arnetheduck, line 1914: it should be this one, with the extra --nonimblepath

arnetheduck: actually, it should also check that if nimble.paths does not exists, it should recommend running `nimble setup`

